PR #19, opened 2020-04-28
feat: Allow updating group description

sklirg:


src/programs/GroupManager.ts
```diff
@@ -23,7 +23,7 @@ export default async function GroupManager(message: Discord.Message, isConfig: b
         const [action, requestName, ...descriptionWords] = words
         const description = descriptionWords.join(" ");
 
-        if (!action || !(["join", "create", "leave", "search", "delete"].includes(action))) {
+        if (!action || !(["join", "create", "leave", "search", "delete", "update"].includes(action))) {
 
             const helpMessage = `Incorrect syntax, please use the following: \`!group join|leave|create|search|delete\`. If you need additional help, react with 🛠️ below to tag a ${ENGINEER_ROLE_NAME}`
```
geisterfurz007, line 28:
```suggestion
            const helpMessage = `Incorrect syntax, please use the following: \`!group join|leave|create|search|delete|update\`. If you need additional help, react with 🛠️ below to tag a ${ENGINEER_ROLE_NAME}`
```

sklirg:
I'll just merge this now since it doesn't affect any current feature, only adds something new. :shipit:

PR #20, opened 2020-04-28
fix: Make profile use GroupRepository for listing group memberships

sklirg:


sklirg:
I've searched through the repository for more usages of 'groupmanager' now and couldn't find any, so this should be fine now :)

src/programs/ProfileManager.ts
```diff
@@ -63,16 +66,14 @@ const getProfileEmbed = async (member:GuildMember, message: Message): Promise<Me
...
-        if(group.members.find(m => m === member.id)) {
-            groupString = groupString+group.name+", "
-        }
-    })
-    groupString = groupString.substring(0, groupString.length - 2);
-    
+    const groupString = groups
+        .filter(group => group.members.some(m => m.id === member.id))
+        .map(group => group.name)
+        .join(', ');
```
geisterfurz007: Couldn't the filtering and mapping be done through the SQL query immediately? I am not familiar with typeorm or the database you put together but I recon something like this should work as well and might be faster (could be relevant later when the groups are a little fuller than they are currently):

```sql
SELECT Name FROM group WHERE member.id IN group.members;
```
Obviously in typeorm but as said, I am not familiar with it

sklirg: <3

PR #26, opened 2020-05-02
feat: moved deadchat questions to database

geisterfurz007:
This change would get rid of both deadchatQuestions.json and usedDeadchatQuestions.json by moving the storage of the questions and their used state to the database.

### DIFF
--- a/src/programs/Deadchat.ts
+++ b/src/programs/Deadchat.ts
@@ -1,5 +1,4 @@
 import Discord, { TextChannel } from 'discord.js';
-import Tools from '../common/tools';
 import { DeadchatQuestion, DeadchatRepository } from '../entities/Deadchat';
 
 export default async function Deadchat(pMessage: Discord.Message) {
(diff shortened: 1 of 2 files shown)


src/programs/Deadchat.ts
```diff
@@ -26,12 +26,16 @@ export default async function Deadchat(pMessage: Discord.Message) {
...
+    const question: DeadchatQuestion = await deadchatRepo
+        .createQueryBuilder()
+        .select()
+        .limit(1)
+        .andWhere("is_used = false")
+        .andWhere("random() < 0.5") // To get a random-ish question (strongly biased towards the top few questions but good enough I guess)
+        .getOne();
+
+    pMessage.channel.send(question.question);
+    question.isUsed = true;
```
sklirg, line 39: For this to be persisted, it has to be saved back into the repository.

However, if all questions are saved as is_used, there won't be any questions for the query above to select.

PR #30, opened 2020-05-03
feat: more flexible polls

geisterfurz007:
Polls now detect options, falling back to A/B if neither of the two attempts to find intended options brought up at least two.

The first attempt consists of checking for emojis at the beginning of a line. All found emojis are added to the array of reactions and returned.

The second attempt looks for capital letters followed by punctuation (to avoid the start of the question for example). Letters are collected, converted to regional indicators and reacted with.

ToDo: In case someone tries to cheat the bot by using the same letter or reaction twice, an error will be thrown (couldn't do this yet because reply is messing around and I don't know why).

### DIFF
--- a/src/events/MessageManager.ts
+++ b/src/events/MessageManager.ts
@@ -70,7 +70,9 @@ class MessageManager {
                 break;
 
             case "polls":
-                Poll(this.message);
+                if (this.message.author.id !== bot.user.id) {
+                    Poll(this.message);
+                }
                 // this.message.react('🇦').then(() => this.message.react('🅱️'))
                 break;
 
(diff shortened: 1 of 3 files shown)


src/events/MessageManager.ts
```diff
@@ -70,8 +70,10 @@ class MessageManager {
                 break;
 
             case "polls":
-
-                this.message.react('🇦').then(() => this.message.react('🅱️'))
+                if (this.message.author.id !== bot.user.id) {
+                    Poll(this.message);
+                }
+                // this.message.react('🇦').then(() => this.message.react('🅱️'))
```
sklirg, line 76: Maybe drop this log?

src/programs/Poll.ts
```diff
@@ -0,0 +1,74 @@
...
+    return defaultOptions;
+}
+
+export default async function Poll(pMessage: Discord.Message) {
+    const options = getOptions(pMessage.content);
+    const uniqueOptions = options.filter((value, index, self) => self.indexOf(value) === index);
+
+    if (uniqueOptions.length < 2) {
+        pMessage.reply("More than one distinct option is required!").then(message => {
+            // message.delete({timeout: 10000});
```
sklirg, line 65: Maybe drop this comment?

PR #31, opened 2020-05-03
fix: UM flag didn't resolve to US (+ BV -> Norway)

geisterfurz007:
The flag made from regional indicators U and M looks indentical to the US flag (see https://emojipedia.org/flag-us-outlying-islands/) which has caught plenty people from the US now.

This PR adds an entry to the flag emoji collection that lets the UM flag resolve to the US in the same way the actual US flag does.

### DIFF
--- a/src/collections/flagEmojis.ts
+++ b/src/collections/flagEmojis.ts
@@ -260,8 +260,8 @@ export const countries: Country[] = [
     "code": "BV",
     "emoji": "🇧🇻",
     "unicode": "U+1F1E7 U+1F1FB",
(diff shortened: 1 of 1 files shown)


sklirg:
Do we want to do this for any similar flags? :flag_bv: looks the same as :flag_no:.

I think yes, since users of emoji keyboards can't easily spot what flag it is, and they will look exactly the same.

geisterfurz007:
Yeah, probably. With Bouvet Islands and Norway it's not too bad because they are at least spaced out by the alphabet while UM and US are close together. I can add BV tho in a second!

PR #38, opened 2020-05-07
fix: timezone evaluation for countries starting with 'the'

geisterfurz007:
Fixes a bug for people adding their birthday with a country role that starts with `I'm from the` (with the `the` being the issue):

grafik

grafik


geisterfurz007:
Sorry about formatting; I got Format on Save turned on in VSCode for all the other projects

PR #39, opened 2020-05-08
Feat/group toggle db

sklirg:


src/events/ReactionAdd.ts
```diff
@@ -71,8 +68,45 @@ class ReactionAdd {
...
+        const toggle = await channelToggleRepository.findOne({
+            where: {
+                emoji: this.reaction,
+                message: this.messageId,
+            }
+        });
+
+        if (toggle !== undefined) {
+            const channel = this.guild.channels.cache.find(channel => channel.id === toggle.channel);
+            await channel.updateOverwrite(this.user.id, {
```
geisterfurz007, line 105: Not sure what the plans are for channels in the future but in case for deleted channels (because of lack of participation or because a channel was split up to smaller ones) this would kill the bot.
Checking if `channel` is defined and in case it's not logging to get moderator attention to the issue would make things more future proof (in case channels are likely to be deleted).

src/programs/GroupManager.ts
```diff
@@ -101,34 +103,79 @@ export default async function GroupManager(message: Discord.Message, isConfig: b
...
+        return
+    }
+
+    if (reactionMessage.channel !== null) {
+        await backfillReactions(reactionMessage.id, reactionMessage.channel, message.guild);
     }
-    
 }
 
+export async function backfillReactions(messageId: string, channelId: string, guild: Guild): Promise<boolean> {
```
geisterfurz007, line 156:
```suggestion
export async function backfillReactions(messageId: string, channelId: string, guild: Guild) {
```

The return value isn't used anywhere

src/programs/GroupManager.ts
```diff
@@ -101,34 +103,79 @@ export default async function GroupManager(message: Discord.Message, isConfig: b
...
+        where: {
+            message: messageId,
+        }
+    });
+
+    // Only add missing reactions
+    toggles
+        .filter(t => !reactionDiscordMessage.reactions.cache.has(t.emoji))
+        .forEach(toggle => reactionDiscordMessage.react(toggle.emoji));
+    return true;
```
geisterfurz007, line 177:
```suggestion
```

Return value isn't used, see other comment.

PR #42, opened 2020-05-13
Allow people to have best friends, significant others, etc 

shagunmistry:
I saw this https://top.gg/bot/marriagebot and thought it'd be a pretty cool feature to have and it'd increase engagement with users.

### DIFF
--- a/src/entities/Family.ts
+++ b/src/entities/Family.ts
@@ -17,6 +17,9 @@ export class Family {
   @Column("text", { nullable: true })
   dad: string;
 
+  @Column("text", { nullable: true })
+  bestfriend: string;
+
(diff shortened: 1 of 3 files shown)


jamie-legg:
This is an interesting feature implementation, i'll throw this branch onto my testing environment and see what the support team think of it. In the past we have denied features like this as we feel like these aren't the sort of connections we want to promote within the Discord, but maybe that ethos has altered somewhat. I'll keep this open and update with details.

shagunmistry:
Yeah, the environment in the server has changed from what I can tell at least. 
I think the feature like best-friends would be fun. The wife-husband feature might be not favored, I guess.

PR #44, opened 2020-05-14
Update bp logic

shagunmistry:


src/programs/BuddyProject.ts
```diff
@@ -1,51 +1,68 @@
 import { db } from "..";
 import { GuildMember, PartialGuildMember } from "discord.js";
 import { BuddyProjectEntryRepository, BuddyProjectEntry } from "../entities/BuddyProjectEntry";
+import { Repository } from "typeorm";
 
 
+const updateDatabseWithQuery = (BE: Repository<BuddyProjectEntry>, memberId: string, buddyId: string, BuddyEntry: BuddyProjectEntry) => {
```
jamie-legg, line 7: `const updateDatabseWithQuery` -> `const updateDatabaseWithQuery`

shagunmistry:
@jamie-legg made the changes
also, included logic to send DM to the matched buddy

src/programs/BuddyProject.ts
```diff
@@ -1,51 +1,104 @@
...
-import { BuddyProjectEntryRepository, BuddyProjectEntry } from "../entities/BuddyProjectEntry";
+import { GuildMember, PartialGuildMember, Client } from "discord.js";
+import {
+  BuddyProjectEntryRepository,
+  BuddyProjectEntry,
+} from "../entities/BuddyProjectEntry";
+import { Repository } from "typeorm";
 
+const updateDatabaseWithQuery = (
+  BE: Repository<BuddyProjectEntry>,
```
sklirg, line 10: Please use readable variable names. It's a lot easier to read and follow the code if the variable name is descriptive. There's no reason to keep the variable names extremely short.

src/programs/BuddyProject.ts
```diff
@@ -1,51 +1,104 @@
...
+  memberId: string,
+  buddyId: string,
+  BuddyEntry: BuddyProjectEntry
+) => {
+  BE.createQueryBuilder()
```
sklirg, line 15: You want to do these two queries in a single transaction, because if one of them fails the database will be inconsistent. You should also handle an error occurring here.

src/programs/BuddyProject.ts
```diff
@@ -1,51 +1,104 @@
...
+    dmChannel.send(successMessage);
 
-    }
-    dmChannel.send(hasEntered.matched ?
-      `It looks like I already found you a match! Did <@${hasEntered.buddy_id}> stopped replying? :grin:`:
-      "Hey there, it looks like you just tried to sign up to the Buddy Project again, no need to do that, you're already registered!" 
-      )
-  }
-  else {
+    if (new Date().getDate() >= 15) {
```
sklirg, line 56: Not really sure what this is supposed to accomplish. Add a comment or add the "15" as a named constant which can be used to infer the use of this.

If it is to disable registration until a specific point in time, I'd suggest creating a _feature toggle_ instead, which makes us able to turn the feature on/off ourselves, rather than having a specific time for it.

src/programs/BuddyProject.ts
```diff
@@ -1,51 +1,104 @@
...
+      // Find matches of the opposite group (aka newsletter group if user is of discord group)
+      const potentialMatches = await buddyEntries.find({
+        where: { discord_user: !discord_user, matched: false },
+      });
+      if (potentialMatches.length > 0) {
```
sklirg, line 61: If you get more than 0 matches, you find some other match instead? I'm not sure I follow.

src/programs/BuddyProject.ts
```diff
@@ -1,51 +1,104 @@
...
+        await buddyEntries
+          .findOne({ where: { discord_user: !discord_user } })
```
sklirg, line 63: This will most likely pick the first signed up user not matched to a buddy. Do we want a selection/randomization algorithm or just the first available?

src/programs/BuddyProject.ts
```diff
@@ -1,51 +1,104 @@
...
+          .then((res) => {
+            updateDatabaseWithQuery(
```
sklirg, line 65: It's a lot easier to read how this function works if it takes an object as a parameter instead of a repository, two strings and an object. Not to mention the order of arguments. :)

src/programs/BuddyProject.ts
```diff
@@ -1,51 +1,104 @@
...
+      }
 
-    const successMessage = "Woo! You just signed up to the buddy project, exciting right? I'll message you again momentarily with your buddy and what you need to do next!";
-    dmChannel.send(successMessage)
-   
-    const potentialMatch = await buddyEntries.findOne({where: { discord_user: !discord_user}})
-
-    if(potentialMatch) {
-      dmChannel.send("Here is your match:" + potentialMatch.user_id);
+      // If the opposite group didn't have any matches, find one from own group
```
sklirg, line 82: What if there is no matches in own group either?

src/programs/BuddyProject.ts
```diff
@@ -1,51 +1,104 @@
...
+        .findOne({ where: { discord_user: discord_user, matched: false } })
+        .then((res) => {
+          updateDatabaseWithQuery(
+            buddyEntries,
+            member.id,
+            res.user_id,
+            BuddyEntry
+          );
 
+          new Client().users.fetch(res.buddy_id).then((mm) =>
```
sklirg, line 93: Is there a specific reason for new-ing up a client rather than reusing the one created in "index"?

src/programs/BuddyProject.ts
shagunmistry: I gotta figure out how to do transactions, haven't had a chance to look at that part of their documentation yet. If you already know how, please let me know or just push it up!

src/programs/BuddyProject.ts
shagunmistry: Ah yeah, good catch. That'll avoid doing that extra call

PR #48, opened 2020-05-14
Update BuddyProject logic based on PR comments

shagunmistry:


src/programs/BuddyProject.ts
```diff
@@ -59,44 +65,48 @@ export async function BuddyProjectSignup(
...
-            res.user_id,
+            finalMatch.user_id,
             BuddyEntry
           );
-
-          new Client().users.fetch(res.buddy_id).then((mm) =>
-            mm.createDM().then((buddyDm) => {
-              buddyDm.send("Here is your match: " + "<@" + member.id + ">");
-            })
+          const buddyDmClient = await new Client().users.fetch(
```
geisterfurz007, line 76: You can `import bot from "../index";` to avoid having to create a new one each time someone signs up for the buddy project.

```suggestion
          const buddyDmClient = bot.users.fetch(
```

src/programs/BuddyProject.ts
```diff
@@ -59,44 +65,48 @@ export async function BuddyProjectSignup(
...
+            finalMatch.user_id
           );
-          dmChannel.send("Here is your match:" + "<@" + res.user_id + ">");
+          buddyDmClient.send(`Here is your match: <@${member.id}>`);
+          dmChannel.send(`Here is your match: <@${finalMatch.user_id}> !`);
+          return null;
```
geisterfurz007, line 81: I don't think there is a reason to return null here.

```suggestion
          return;
```

src/programs/BuddyProject.ts
```diff
@@ -59,44 +65,48 @@ export async function BuddyProjectSignup(
...
+          where: { discord_user: discord_user, matched: false },
         });
+        updateDatabaseWithQuery(
+          buddyEntries,
+          member.id,
+          finalMatch.user_id,
+          BuddyEntry
+        );
+
+        const buddyDmClient = await new Client().users.fetch(
```
geisterfurz007, line 102: See comment above

```suggestion
        const buddyDmClient = bot.users.fetch(
```

src/programs/BuddyProject.ts
```diff
@@ -7,21 +7,27 @@ import {
...
 const updateDatabaseWithQuery = (
-  BE: Repository<BuddyProjectEntry>,
+  BuddyEntryRepo: Repository<BuddyProjectEntry>,
   memberId: string,
   buddyId: string,
   BuddyEntry: BuddyProjectEntry
 ) => {
-  BE.createQueryBuilder()
+  BuddyEntryRepo.createQueryBuilder()
     .update(BuddyEntry)
```
geisterfurz007, line 16: I am not really familiar with TypeORM but I think the BuddyEntry isn't needed here. In fact I suspect this to be the reason that people aren't actually being matched (however I would like @jamie-legg or @sklirg to double check this):

grafik


```suggestion
    .update()
```

src/programs/BuddyProject.ts
```diff
@@ -7,21 +7,27 @@ import {
...
     .set({ matched: true, buddy_id: buddyId })
     .where("used_id = :member_id", { member_id: memberId })
-    .execute();
-  BE.createQueryBuilder()
+    .execute()
+    .catch((err) =>
+      console.log("There was an error updating member entry: ", err)
+    );
+  BuddyEntryRepo.createQueryBuilder()
     .update(BuddyEntry)
```
geisterfurz007, line 24: Same as above

```suggestion
    .update()
```

src/programs/BuddyProject.ts
```diff
@@ -7,21 +7,27 @@ import {
...
+  BuddyEntryRepo: Repository<BuddyProjectEntry>,
   memberId: string,
   buddyId: string,
   BuddyEntry: BuddyProjectEntry
 ) => {
-  BE.createQueryBuilder()
+  BuddyEntryRepo.createQueryBuilder()
     .update(BuddyEntry)
     .set({ matched: true, buddy_id: buddyId })
     .where("used_id = :member_id", { member_id: memberId })
```
geisterfurz007, line 18: That is a big problem.

```suggestion
    .where("user_id = :member_id", { member_id: memberId })
```

src/programs/BuddyProject.ts
```diff
@@ -7,21 +7,27 @@ import {
...
-    .execute();
-  BE.createQueryBuilder()
+    .execute()
+    .catch((err) =>
+      console.log("There was an error updating member entry: ", err)
+    );
+  BuddyEntryRepo.createQueryBuilder()
     .update(BuddyEntry)
     .set({ matched: true, buddy_id: memberId })
     .where("used_id = :member_id", { member_id: buddyId })
```
geisterfurz007, line 26: See above

```suggestion
    .where("user_id = :member_id", { member_id: buddyId })
```

shagunmistry:
@geisterfurz007 please review again when possible, made changes!

PR #51, opened 2020-05-16
feat: auto assign and remove unassigned for new members

geisterfurz007:
Potential quick win:

New users get unassigned on join and have it removed when they receive a country role from the bot.
Especially with a lot of new users coming in, it's probably a good idea to have some automated way of doing this.

Potential alternative: Scratch Unassigned all together and use `@everyone` in flag-drop?

Opening this as draft before it's merged and then not needed. I am going to open or close once results on the above question are in!

### DIFF
--- a/src/events/MemberJoin.ts
+++ b/src/events/MemberJoin.ts
@@ -8,6 +8,9 @@ class MemberJoin {
     bot: Discord.Client;
 
     constructor(member: GuildMember | PartialGuildMember) {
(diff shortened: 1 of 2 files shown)


geisterfurz007:
Looks like `@everyone` works just fine and while Unassigned is still a thing, I am going to close this and delete the branch. I doubt this will become relevant again in the future and if so, the 5 lines shouldn't be that hard to reproduce.

PR #53, opened 2020-05-17
fix: ignore group pings that are in a blockquote

geisterfurz007:
I *think* this works, in tests it looks good but it is 4 in the morning and I'd be happy if someone could double check this.

Background: It happens that messages are quotes that contain a group ping. Those shouldn't cause a new group ping:

grafik

This PR (somewhat lazily) fixes this by checking if there is any group ping outside quotes. That still allows this behaviour (the first one actually tries to ping group thisfails instead of something):

grafik

But that is a problem for another PR, I guess.

### DIFF
--- a/src/programs/GroupManager.ts
+++ b/src/programs/GroupManager.ts
@@ -79,13 +79,15 @@ export default async function GroupManager(message: Discord.Message, isConfig: b
     }
 
     else {
-        // Matches if there is at least one line containing @group that doesn't start with a >
-        const match = /^(?!>).*@group/gm;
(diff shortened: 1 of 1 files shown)


sklirg:
Would it be more readable to split the message on newlines and filter away any line starting with `>` instead of the regex match?

geisterfurz007:
Yup; will change in a few!

geisterfurz007:
I updated the code according to the suggestions. That also made it a lot easier to actually fully ignore quoted group pings so this is how it looks now:

grafik

PR #62, opened 2020-05-27
Fixes for Oman - Romania issues

geisterfurz007:
This PR fixes two issues with Oman and Romania in where are you from:

 - The flag of Oman got the user the role for Romania
 - Typing out Romania got two found country entries (Romania and Oman) because of the way how countries are checked.

### DIFF
--- a/src/programs/WhereAreYouFromManager.spec.ts
+++ b/src/programs/WhereAreYouFromManager.spec.ts
@@ -1,0 +1,62 @@
+import { getCountriesFromMessage, getRoleForCountry } from './WhereAreYouFromManager';
+import bot from '../index';
+import { GUILD_ID } from '../const';
+
+// message, countryName, roleName
+type Testcase = [string, string, string];
+
+const cases: Array<Testcase> = [
+  ["🇴🇲", "Oman", "I'm from Oman!"],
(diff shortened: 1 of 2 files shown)


sklirg:
Sorry about the slow review!! 😞

I'd like to see some tests for this. It should be easy to pick out some of the problematic cases we've seen, compile them into a list of cases such as 
```js
const cases = [
  "🇴🇲",
  "🇷🇴",
  "romania",
  "norway",
];
```
and simply run this function for each of those entries. Look at BirthdayManager.spec.ts for inspiration.

geisterfurz007:
Good idea, thanks for the suggestion! I put together a few testcases like you suggested which can be run with `npx ts-node src\programs\WhereAreYouFromManager.spec.ts`.
As Oman didn't exist as role when the template was created, you will likely have to create the role, I am assuming it does exist on the live server.

jamie-legg:
> Good idea, thanks for the suggestion! I put together a few testcases like you suggested which can be run with `npx ts-node src\programs\WhereAreYouFromManager.spec.ts`.
> As Oman didn't exist as role when the template was created, you will likely have to create the role, I am assuming it does exist on the live server.

Role is there, I liked your original implementation but agree with adding tests.

PR #65, opened 2020-06-02
Feat/buddy project ghosts

geisterfurz007:
This PR aims to provide a way for users who were matched in the buddy project to report that they have not heard back from their buddies and would like to get matched again.

This is achieved by marking users as ghosted in the database and running a script periodically to check for users who were reported as ghosted more than a week ago. Those users and their buddies are unmatched and the user who reported being ghosted is matched with a new user.

To provide a way to prevent people just yeeting out their old buddy in favour of a new one, a message is sent to the buddy asking them to confirm that they are still using Discord and are willing to message their buddy within a week by clicking a reaction below the message. When they click it, their buddy (the user who originally reported being ghosted) is messaged, letting them know about it and the database entry is changed to reflect that the user should not be ghosted anymore.

I would appreciate some input on this; I made this a draft before it's merged accidentally potentially breaking a lot of stuff in case I messed up somewhere!


src/programs/BuddyProjectGhost.ts
```diff
@@ -0,0 +1,69 @@
...
+  const buddy = guild.member(entry.buddy_id);
+
+  if (!buddy) {
+    userDm.send("Huh... Looks like I couldn't find your buddy! I reported that to the support team and have them check it. Expect them to get back to you in a bit!");
+    textLog(`Heyyo! <@${user.id}> reported that their buddy <@${buddy.id}> hasn't replied yet, however I could not find that user and couldn't create a DM for that reason. Could one of you try to contact them by DM?`);
+    return;
+  }
+
+  const buddyDm = await buddy.createDM();
+  const buddyMessage = await buddyDm.send(`Hey there! You signed up to the buddy project and got matched with <@${user.id}> (${user.username}#${user.tag}) but never responded to them. In order to not get unmatched, click on the checkmark below this message within the next seven days and send them a message :)`);
```
shagunmistry, line 42: Do we want to make the user wait 7 days again? Maybe allow 48 hours to respond?

geisterfurz007: Personal opinion ahead:

I think one week is a good timespan because it ensures there is a weekend in it. There are plenty people who don't open Discord during the week.

Also in earlier iterations of the buddy project there was no such mechanism and people who had been ghosted were just unlucky. Waiting a week is better than what is currently available, was available in older iterations of BP and hopefully not too much to ask for (although I wouldn't be surprised to see the impatient bunch on the server complain about that as well).

PR #75, opened 2020-06-16
Fix myself error

shagunmistry:


geisterfurz007:
Would be great if you could run `npm run lint:fix` to get the CI green :)

Another concern: In the near future we will have a pretty clean db in terms of unmatched users (at least following the current plans). I would like to see the filter before the condition; otherwise if only the user who just signed up is in the database, this will cause an error because there is no result to the find.

PR #76, opened 2020-06-17
feat: updated !clean to match unmatched members

geisterfurz007:
Quick implementation of what was requested by Léa and originally planned for the clean command.
Untested so draft, I hope I have a shorter day tomorrow and can give this a trial run locally. If someone else gave this a shot and it worked as intended, let me know and I will open this for merging.

I also had another idea for an additional note in the bottom letting the mod know whether one user was left because of an odd number of registered users in the database at point of execution.

Will also take care of the failing CI tomorrow; I have to change some stuff about my local git setup to avoid wrong lineendings making prettier unusable for me currently.


geisterfurz007:
CI will have to wait until tomorrow, I should finally have more time tomorrow and over the weekend.

geisterfurz007:
@jamie-legg I am currently in the process of cleaning up old branches of mine and found this again. Given that the project restarted, that this branch is an absolute mess and that Léa seems to be able to match people faster than I can write this command, should I just close this PR and drop the branch or is there any reason to keep it?

geisterfurz007:
No response but doesn't look like it's needed anymore so closing

PR #77, opened 2020-06-18
Get all unmatched ids

shagunmistry:


src/programs/BuddyProject.ts
```diff
@@ -381,3 +381,16 @@ export const cleanEntries = async (guild: Guild) => {
...
+  const entries = await BuddyProjectEntryRepository();
+  const all = await entries.find({ where: { matched: false } });
+
+  const outputChannel = guild.channels.cache.find(
+    (c) => c.name == "buddy-project-matches"
+  ) as TextChannel;
+  const returnString = "";
+  all.forEach((each) => returnString + ` , ${each.user_id}`);
+  let outputText = `All unmatched members: ${returnString}`;
+  outputChannel.send(outputText);
```
geisterfurz007, line 395:
```suggestion
  outputChannel.send(outputText, { split: true });
```
I don't know how many characters this is going to be but I wouldn't be surprised if it was more than 2k. Better be safe than sorry.

src/programs/BuddyProject.ts
```diff
@@ -381,3 +381,16 @@ export const cleanEntries = async (guild: Guild) => {
...
+
+export const checkAllEntries = async (guild: Guild) => {
+  const entries = await BuddyProjectEntryRepository();
+  const all = await entries.find({ where: { matched: false } });
+
+  const outputChannel = guild.channels.cache.find(
+    (c) => c.name == "buddy-project-matches"
+  ) as TextChannel;
+  const returnString = "";
+  all.forEach((each) => returnString + ` , ${each.user_id}`);
```
geisterfurz007, line 393:
```suggestion
  const returnString = all.map((each) => each.user_id).join(", ");
```
This is just a nice little touch; imo a little easier to read and doesn't produce a leading ` , `. Tbf, it is not really relevant because this will only be usable by us but figured, I'd still address it.

shagunmistry: Fixed

PR #80, opened 2020-06-21
Feat: MemberLeave

shagunmistry:
Remove the member from as many DB as possible
Let me know if I missed any DBs

### DIFF
--- a/src/events/MemberLeave.ts
+++ b/src/events/MemberLeave.ts
@@ -9,14 +9,10 @@ export class MemberLeave {
   constructor(member: GuildMember | PartialGuildMember) {
     if (member.roles.cache.find((r) => r.name === "Buddy Project 2020")) {
       RemoveFromBuddyProject(member.id);
-      RemoveFromBirthdays(member.id);
-      RemoveFromGroups(member.id);
     }
-    member.createDM().then((dm) => {
-      dm.send(
-        `Hi ${member.displayName}! Sorry to hear that you left the server! :( If you would ever like to join back, just let me know! :slight_smile: `
-      );
-    });
+
+    RemoveFromBirthdays(member.id);
+    RemoveFromGroups(member.id);
   }
 }
 
--- a/src/events/MemberLeave.ts
+++ b/src/events/MemberLeave.ts
@@ -1,0 +1,90 @@
+import {
+    GuildMember,
+    PartialGuildMember,
+} from "discord.js";
+import { BuddyProjectEntryRepository } from "../entities/BuddyProjectEntry";
+import { BirthdayRepository } from "../entities/Birthday";
+import { GroupMember, UserGroupRepository } from '../entities/UserGroup'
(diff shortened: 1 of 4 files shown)


src/events/MemberLeave.ts
```diff
@@ -0,0 +1,93 @@
...
+export class MemberLeave {
+  constructor(member: GuildMember | PartialGuildMember) {
+    if (member.roles.cache.find((r) => r.name === "Buddy Project 2020")) {
+      RemoveFromBuddyProject(member.id);
+      RemoveFromBirthdays(member.id);
+      RemoveFromGroups(member.id);
+    }
+    member.createDM().then((dm) => {
+      dm.send(
+        `Hi ${member.displayName}! Sorry to hear that you left the server! :( If you would ever like to join back, just let me know! :slight_smile: `
```
jamie-legg, line 17: We can't send user's DMs if they have no servers in common. I would opt to remove this functionality altogether.

src/events/MemberLeave.ts
```diff
@@ -0,0 +1,93 @@
...
+import { BirthdayRepository } from "../entities/Birthday";
+import { GroupMember, UserGroupRepository } from "../entities/UserGroup";
+import { textLog } from "../common/moderator";
+import { ILike } from "../lib/typeormILIKE";
+
+export class MemberLeave {
+  constructor(member: GuildMember | PartialGuildMember) {
+    if (member.roles.cache.find((r) => r.name === "Buddy Project 2020")) {
+      RemoveFromBuddyProject(member.id);
+      RemoveFromBirthdays(member.id);
```
jamie-legg, line 12: Why are we only doing this if they have the buddy project role? Shouldn't birthday and groups be done regardless of if they were in the buddy project or not?

src/index.ts
```diff
@@ -1,37 +1,38 @@
-import {Client, Channel, Emoji, Guild, GuildMember, PartialGuildMember, Message, User, PartialUser, Collection, Role, TextChannel, Snowflake, MessageReaction, Speaking, PartialMessage, Presence, VoiceState,} from 'discord.js';
+import { Client, Channel, Emoji, Guild, GuildMember, PartialGuildMember, Message, User, PartialUser, Collection, Role, TextChannel, Snowflake, MessageReaction, Speaking, PartialMessage, Presence, VoiceState, } from 'discord.js';
```
jamie-legg, line 1: Unnecessary whitespace
```suggestion
import { Client, Channel, Emoji, Guild, GuildMember, PartialGuildMember, Message, User, PartialUser, Collection, Role, TextChannel, Snowflake, MessageReaction, Speaking, PartialMessage, Presence, VoiceState } from 'discord.js';
```

src/events/MemberLeave.ts
shagunmistry: yep, makes sense

PR #87, opened 2020-06-23
update !buddy command

shagunmistry:


src/programs/BuddyProjectManager.ts
```diff
@@ -62,8 +62,16 @@ export default async function BuddyProjectManager(
         .then((sentMsg) => sentMsg.react("👻"));
       break;
     case "buddy":
-      const buddy = await getBuddyId(message.author);
-      message.reply( buddy ? `Your buddy is <@${await getBuddyId(message.author)}>.` : "You haven't entered the Buddy Project yet! Go to <#701153345620148335> and react on the post to enter. :grin:");
+      const mentionedUser = message.mentions.users.first();
+      let buddy = await getBuddyId(message.author);
+      if (mentionedUser !== undefined) {
+        buddy = await getBuddyId(mentionedUser);
+      }
```
geisterfurz007, line 69:
```suggestion
      let buddy;
      if (mentionedUser) buddy = await getBuddyId(mentionedUser);
      else buddy = getBuddyId(message.author);
```
We don't have to read from the database if we don't need the data.

src/programs/BuddyProjectManager.ts
```diff
@@ -62,8 +62,16 @@ export default async function BuddyProjectManager(
...
+      message.reply(
+        buddy
+          ? `Your buddy is <@${buddy}>.`
+          : "You haven't entered the Buddy Project yet! Go to <#701153345620148335> and react on the post to enter. :grin:"
```
geisterfurz007, line 73: These probably should be adapted as well.

shagunmistry:
Not needed anymore so closing

PR #91, opened 2020-06-28
fix: allow group pings in voice-chat

geisterfurz007:


geisterfurz007:
Already on master, yeet

PR #100, opened 2020-07-04
fix: durations in buddy project ghost process

geisterfurz007:
I accidentally made it so that people only have to be matched for three days before being able to report themselves as ghosted. That should have remained at one week. The three days were only the timespan between report and getting a new buddy.


geisterfurz007:
Also happy #100

PR #103, opened 2020-07-05
fix: broken group ping when used immediately after a linebreak

geisterfurz007:
Because messages were only split on spaces, using a group ping immediately after a linebreak would leave the entry in the words array as `something\n@group`.

This PR changes splitting to whitespace in general.


jamie-legg:
Great catch on this bug, I would never have suspected this.

PR #106, opened 2020-07-17
finishing touches to voice on demand remake

geisterfurz007:
- Cleans up empty channels if they went empty during a restart
 - No duplicate timeouts per channel that made them get deleted one minute after the last person left
    (Last person leaves, comes back after 50 seconds, leaves 5 seconds later, channel was deleted 5 seconds after that)
 - No pinging the Yes Theory role anymore when someone uses it without the role (that could have been a lot of fun)
 - Changed max limit from 10 to 30
 - False command calls and error messages are properly cleaned up
 - Simplified permissions

### DIFF
--- a/src/programs/VoiceOnDemand.ts
+++ b/src/programs/VoiceOnDemand.ts
@@ -4,11 +4,15 @@ import {
   VoiceState,
   VoiceChannel,
   Permissions,
+  Client,
 } from "discord.js";
 
 import { hasRole } from "../common/moderator";
+import state from "../common/state";
+import { GUILD_ID } from "../const";
 import { VoiceOnDemandRepository } from "../entities/VoiceOnDemandMapping";
 
+const emptyTime = 60000;
 const getChannelName = (m: GuildMember) => `• 🔈 ${m.displayName}'s Room`;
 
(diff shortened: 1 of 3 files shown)


src/programs/VoiceOnDemand.ts
```diff
@@ -194,3 +216,10 @@ const deleteIfEmpty = async (channel: VoiceChannel) => {
     repo.delete({ channelId: channel.id });
   }
 };
+
+const error = (message: Message, reply: string) => {
+  message.reply(reply).then((msg) => {
+    message.delete();
+    msg.delete({ timeout: 10000 });
+  });
+};
```
shagunmistry, line 225: This should be a global function right? If so, lets move it into a common file

geisterfurz007: Yeah, I was wondering about that. I can do another PR in the future to move the function to src/common/tools and convert all existing stuff to that but I would like to keep it out of this one.

jamie-legg: Could definitely be one, we'll see if we have further need in the future for a global error handler

PR #114, opened 2020-08-24
Group ping for fourth chat and some logic cleanup

shagunmistry:


src/programs/GroupManager.ts
```diff
@@ -638,3 +635,27 @@ const groupInteractionAndReport = async (
...
+  if (!isTextChannel(channel)) return;
+
+  const allowedCategories = ["hobbies", "gaming"];
+  const allowedChannels = [
+    "449984633908625409",
+    "623565093166252052",
+    "508918747533410304",
+    "450187015221280769",
+    "747189756302983199",
+  ];
```
geisterfurz007, line 651: Not sure how I feel about having one as names and one as IDs. Drawbacks of the IDs is losing information on the human readable name (if we had to delete #chat-too from the list for example, it would take a while to find out which one it is from that list).
Maybe make everything IDs and comment the channel name behind it or something like that is a good compromise?

geisterfurz007:
Also *big* thanks for moving the validation logic out of the MessageManager!

src/programs/GroupManager.ts
shagunmistry: sure, added the comments

PR #116, opened 2020-08-25
fix for someone

shagunmistry:


src/programs/Someone.ts
```diff
@@ -66,6 +66,7 @@ const sendMessage = async (
   const webhook = await channel.createWebhook(author.displayName, {
     avatar: author.user.avatarURL(),
   });
+  await webhook.send(`<@${target.id}> ${question}`);
```
shagunmistry, line 69: this is the fix. Not sure why it wasn't there before. Unless I am misunderstanding something

geisterfurz007: Good catch! Looks like it went missing in the cleanup of unused variables: https://github.com/Yes-Theory-Fam/yesbot-ts/commit/1fe65c3aaf9ab329d7a37dcd99ccdd17ec1628d8#diff-a6400de07507528bce163e52f7f032a9L69

PR #128, opened 2020-09-05
Feat/voice chat transfer

geisterfurz007:
This PR introduces a new feature connected to !voice: Automatic transfer to another user when the original owner leaves.

**Why?**
Imagine someone creates a room, others join, first guy has to leave and the others stick around. Now the limit of the room is stuck to what it is set to because noone else can control it (except for mods which brings us back to the original issue :smile:).

**How?**
When the original owner leaves a timer for one minute is set. If the room isn't empty after a minute, the bot pings all members in the voice channel asking one of them to react with :point_up:. The first person to do so will become the new owner of the room (the name changes and they are able to use !voice limit; emoji stays the same for now). This also allows the original owner to create a new room if they want to as they don't own one anymore.
If noone reacts within a minute, the room is deleted.

### DIFF
--- a/src/programs/VoiceOnDemand.ts
+++ b/src/programs/VoiceOnDemand.ts
@@ -182,7 +182,15 @@ export const voiceOnDemandPermissions = async (
 
   const { guild } = channel;
 
+  const timeoutRole = guild.roles.cache.find(
+    (role) => role.name.toLowerCase() === "time out"
+  );
+
   channel.overwritePermissions([
+    {
+      id: timeoutRole.id,
+      deny: [Permissions.FLAGS.CONNECT],
+    },
     {
       id: guild.roles.everyone,
       allow: [Permissions.FLAGS.STREAM],
--- a/src/programs/VoiceOnDemand.ts
+++ b/src/programs/VoiceOnDemand.ts
@@ -1,6 +1,5 @@
 import {
   Client,
-  Emoji,
   GuildMember,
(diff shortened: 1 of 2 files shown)


src/programs/VoiceOnDemand.ts
```diff
@@ -170,7 +175,15 @@ export const voiceOnDemandPermissions = async (
 
   const { guild } = channel;
 
+  const timeoutRole = guild.roles.cache.find(
+    (role) => role.name.toLowerCase() === "time out"
+  );
```
shagunmistry, line 180: Can we make this a re-usable method in a common file? A method that takes in the role name and returns it if it finds it

geisterfurz007: Yup, going to add a larger refactoring commit to this in a bit (also going to do [this one](https://github.com/Yes-Theory-Fam/yesbot-ts/pull/106#discussion_r458341243)), making use of them as needed.

geisterfurz007: Seems I accidentally did this on master :trollface:

Relevant commit: https://github.com/Yes-Theory-Fam/yesbot-ts/commit/60ad73c24d8e6713843638218a223a117b19b402

findtravishere (approved):

shagunmistry (approved):

dimapanf (approved):

PR #129, opened 2020-09-09
Start a global logger + format

shagunmistry:


src/common/CustomMethods.ts
```diff
@@ -37,3 +46,96 @@ export const randomReply = (message: Message) => {
 export const reactWithEmoji = (message: Message, emoji: string) => {
   message.react(emoji);
 };
+
+export const proposeNameChange = async (name: string, botMessage: Message) => {
+  Logger(
+    "CustomMethods",
+    "proposeNameChange",
+    "Name change request sent to mods"
+  );
```
geisterfurz007, line 55: Adding more logging around errors certainly won't harm, I am a little opposed to logging things like this tbh. When something breaks on the server, the first reaction is pretty much 100% going to be connecting to the cloud instance and running journalctl to see what happened. I want to see things that broke and not a log cluttered with things we have other means of output for (in this specific case we know someone wants a name change, because it's sent in the bot-output channel).  
Thus I would like to see this logging bit removed.

src/common/moderator.ts
```diff
@@ -6,8 +6,14 @@ import {
...
 } from "discord.js";
 import { MODERATOR_ROLE_NAME, OUTPUT_CHANNEL_ID, GUILD_ID } from "../const";
+import { Logger } from "./Logger";
 
 export const isAuthorModerator = (message: Message): boolean => {
+  Logger(
+    "moderator",
+    "isAuthorModerator",
+    `Checking if author is mod: '${message.author.id}'`
+  );
```
geisterfurz007, line 16: Same thing as above (this *especially* because it's called a lot).

src/common/moderator.ts
```diff
@@ -19,22 +25,34 @@ export const hasRole = (
   member: GuildMember | PartialGuildMember,
   roleName: string
 ): boolean => {
+  Logger(
+    "moderator",
+    "hasRole",
+    `Checking if user has role: ${roleName} for: '${member.id}'`
+  );
```
geisterfurz007, line 32: Same

src/common/moderator.ts
```diff
@@ -19,22 +25,34 @@ export const hasRole = (
...
 };
 
 export const isRegistered = (
   member: GuildMember | PartialGuildMember
 ): boolean => {
+  Logger(
+    "moderator",
+    "isRegistered",
+    `Getting Registerd info for: '${member.id}'`
+  );
```
geisterfurz007, line 43: Same

src/common/moderator.ts
```diff
@@ -19,22 +25,34 @@ export const hasRole = (
...
   return !!member.roles.cache.find((role) => role.name.startsWith("I'm from "));
 };
 
 export const textLog = (text: string): Promise<Message> => {
+  Logger("moderator", "textLog", `Sending this message: '${text}'`);
```
geisterfurz007, line 48: Same

src/common/moderator.ts
```diff
@@ -19,22 +25,34 @@ export const hasRole = (
...
   const bot = require("..") as Client;
   const outputChannel = <TextChannel>bot.channels.resolve(OUTPUT_CHANNEL_ID);
   return outputChannel.send(text);
 };
 
 export const getMember = (userId: string): GuildMember => {
+  Logger("moderator", "getMember", `Getting member info for: '${userId}'`);
```
geisterfurz007, line 55: Same

src/events/MessageManager.ts
```diff
@@ -248,9 +250,14 @@ class MessageManager {
...
-      this.proposeNameChange(requestedName);
+      proposeNameChange(requestedName, this.message);
       await requestMessage.delete();
-    } catch {
+    } catch (err) {
+      Logger(
+        "MessageManager",
+        "routeDM",
+        `Timed out wait for a reaction: ${err}`
+      );
```
geisterfurz007, line 260: I also wouldn't log this as an error I think. It's more or less intended behaviour, not caused by an error in the code but because we explicitly tell discord.js to throw an error when the time limit is up.

src/programs/BirthdayManager.ts
```diff
@@ -288,6 +313,7 @@ async function getUserTimezone(message: Message): Promise<string> {
       errors: ["time"],
     });
   } catch (err) {
+    Logger("BirthdayManager", "getUserTimezone - awaitReactions", err);
```
geisterfurz007, line 316: Same as above; this error being thrown is pretty much intentional

src/events/MessageManager.ts
```diff
@@ -192,7 +194,7 @@ class MessageManager {
       this.message.content.toLowerCase().includes("abooz") ||
       this.message.content.toLowerCase().includes("mod abuse")
     ) {
-      reactWithEmoji(this.message, ":mod_abooz:");
+      reactWithEmoji(this.message, ":eyes:");
```
geisterfurz007, line 197: This won't work; if a string is passed, it has to be either a unicode character or an ID of a custom emoji.

geisterfurz007 (approved):

geisterfurz007:
:shipit:

PR #131, opened 2020-09-16
at-someone abuse fix

findtravishere:
included regex to prevent @s from being parsed in argument.


findtravishere:
removed regex for discordjs features instead :')

geisterfurz007 (approved):

Issue #143, opened 2020-10-07
Fix some country overrides


geisterfurz007:
This one is similar to #141 and references the same file. In this case we are looking to update existing entries to match the required format/strings that we are using.

As background: When a user drops a flag in the flag-drop channel, the [`flagEmojis.ts`](https://github.com/Yes-Theory-Fam/yesbot-ts/blob/master/src/collections/flagEmojis.ts) are loaded and the emoji is looked up. If an entry is found, the name field of the object is used for a search through the server's roles by searching for a role that starts with `"I'm from"` and ends with `country.name + "!"`.

Several name entries don't match the role names on the server. This is likely not a complete list but one that should suffice for now:

- British Indian Ocean Territory => UK
- Palestinian Territory => Palestine
- Syrian Arab Republic => Syria
- Saint Martin (French Part) => France

geisterfurz007:
Closed in #144

PR #144, opened 2020-10-07
added syria, palestine, BIOT, SM as selectable roles

dimapanf:
changed 4 region names so mods dont get pinged that the role doesn't exist when it does. can update list over time


geisterfurz007 (approved):

geisterfurz007:
Closes #143

PR #159, opened 2020-10-28
Updated prompt - owner left

huffle67:
Self explanatory name, changed the prompt when owner leaves their custom voice channel to say it'll pick a new owner at random instead of it'll delete the channel.


geisterfurz007 (approved):
:shipit:

PR #160, opened 2020-10-31
set chances of ghost appearing to 30%

huffle67:
self explanatory name
if message contains oo, bot will react with ghost 30% of the time


geisterfurz007 (approved):
Marvelous

PR #161, opened 2020-11-01
Feat/voice knock

### LINKED ISSUE #146: !voice knock
It often happens that a user wants to join a !voice created channel that reached its limits. They then ping the owner of the room, that person has to check the limit and manually increment the limit.

The idea is to have a `!voice knock @user` command, that responds with "Let that user in?" with a 👍 and 👎 reactions. When 👍 is clicked, the room limit is increased.

This requires a larger refactoring in the [VoiceOnDemand.ts](https://github.com/Yes-Theory-Fam/yesbot-ts/blob/master/src/programs/VoiceOnDemand.ts) to move `create`, `limit` and `host` to a single "branch" with the common checks (has Yes Theory, has a room, doesn't have a room, etc) and add a new command that looks up the pinged user in the DB, checks for their room, gets the limit and increases it by one **unless that would raise it above 10**.

geisterfurz007:
Closes #146 
This also adds three more commands to !voice:

- up: Increases the limit by one (unless the room is at the max of 10)
- down: Decreases the limit by one (unless the room is at the min of 2)
- shrink: Sets the limit to the amount of members (or 2 if someone is alone in a room)

### DIFF
--- a/src/programs/VoiceOnDemand.ts
+++ b/src/programs/VoiceOnDemand.ts
@@ -51,6 +51,14 @@ export default async function (message: Message) {
     case "limit":
       handleLimitCommand(message);
       break;
+    case "shrink":
+    case "up":
+    case "down":
+      handleLimitUpdateCommand(message, command);
+      break;
+    case "knock":
+      knockOnDemand(message);
+      break;
     case "host":
       changeHostOnDemand(message);
       break;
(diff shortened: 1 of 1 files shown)


shagunmistry (approved):

dimapanf (commented):
maybe change the downvote string to something more positive like "doesnt want more people at the moment".

geisterfurz007:
> maybe change the downvote string to something more positive like "doesnt want more people at the moment".

Good catch; I already forgot again that this was in there. It was something arch wanted to chat about to find something better for it. Eventually they decided to just remove it so if someone doesn't want someone else to join, they just don't do anything.

dimapanf (approved):

PR #163, opened 2020-11-01
added knock knock joke

huffle67:
added who's there when someone types a message that starts with !knock knock (i think, maybe, who knows what i've done)


huffle67:
closed it cause i was supposed to do !voice knock knock instead of just !knock knock but ima redo it so it's all good dw

PR #166, opened 2020-11-02
fix: Proper message when member is null

geisterfurz007:
For some reason guildMember is null in extremely rare cases and I don't know why so we just provide a proper error message until it's figured out.

### DIFF
--- a/src/events/ReactionAdd.ts
+++ b/src/events/ReactionAdd.ts
@@ -89,8 +89,10 @@ class ReactionAdd {
         reaction: this.reaction,
       },
     });
-    reactRoleObjects.forEach((reactionRole) => {
-      const guildMember = this.guild.member(this.user);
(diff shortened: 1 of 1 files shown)


dimapanf (commented):
what can we do if discord finds them as null? maybe have them retry the command OR we could forward a msg straight to bot-commands saying a user appeared as null so can  you please manually set their colour

geisterfurz007:
Good question! Forced me to think again about the underlying issue which got me to a much more reasonable solution. BaseManager#resolve only uses the cache to find values for IDs to remain fast and synchronous; I trashed all the DM stuff and now actively fetch the member from the API when the cache doesn't have the requested entry.

shagunmistry (approved):

dimapanf (approved):

PR #167, opened 2020-11-03
Let users tag in a roast

shagunmistry:


src/common/CustomMethods.ts
```diff
@@ -48,6 +48,9 @@ export const reactWithEmoji = (message: Message, emoji: string) => {
 };
 
 export const abuseMe = (message: Message) => {
+  const taggedUser = message.mentions.users
+    ? message.mentions.users.array()[0]
+    : null;
```
geisterfurz007, line 53:
```suggestion
  const taggedUser = message.mentions.users?.first();
```

src/common/CustomMethods.ts
```diff
@@ -57,7 +60,18 @@ export const abuseMe = (message: Message) => {
...
+    ? `<@${
+        taggedUser.id
+      }> *\`${message.content.toString()}\`* translated to English means *${
+        replies[Math.floor(Math.random() * replies.length)]
+      }*`
+    : `<@${
+        message.author.id
+      }> *\`${message.content.toString()}\`* translated to English means *${
+        replies[Math.floor(Math.random() * replies.length)]
+      }*`;
```
geisterfurz007, line 73: This is heavily duplicated; probably worth assigning the ID of the tagged user first, then put it to the string.

geisterfurz007 (changes requested):
And please squash for a clearer commit message

src/common/CustomMethods.ts
```diff
@@ -48,16 +48,37 @@ export const reactWithEmoji = (message: Message, emoji: string) => {
 };
 
 export const abuseMe = (message: Message) => {
+  const taggedUser = message.mentions.users?.array()[0];
```
geisterfurz007, line 51:
```suggestion
  const taggedUser = message.mentions.users?.first();
```

geisterfurz007 (changes requested):
'tis broken:

grafik

`message.cleanContent` might work in place of `message.content`.

geisterfurz007 (approved):

PR #174, opened 2020-11-07
Fix/issues

### LINKED ISSUE #149: Fix output bug occuring when noone claims a room but noone is in there
When a !voice created room is left by the owner of it, a request for someone still in the room is sent after a minute. If this request is left unanswered, the bot chooses someone randomly.   
This leaves opportunity for the last person to leave within the minute after the request and this happens:

grafik

This should be fixed; one minute after the request is sent, the bot should check whether anyone is in the room and if that's not the case, not output anything.

The code can be found in [VoiceOnDemand.ts#requestOwnershipTransfer](https://github.com/Yes-Theory-Fam/yesbot-ts/blob/master/src/programs/VoiceOnDemand.ts#L348). While at it, please uncomment the line removing the request after one minute.

### LINKED ISSUE #165: Fix the algorithm that finds an online member with the seek discomfort role
The [`getTarget`](https://github.com/Yes-Theory-Fam/yesbot-ts/blob/master/src/programs/Someone.ts#L101) function is not cleanly implemented and deserves some maintenance.

With the increase of members with the Seek Discomfort role, the chance of finding someone online with the role has decreased and the error message of not finding a user has shown more and more.

Instead of looping 100 times, each time picking a new random member, first a collection of members should be created depending on whether they should be online or not (using something like `Collection#filter`). Then pick a random one of those.

This should both create cleaner code and make finding an online member with the role a lot more reliable.

### LINKED ISSUE #169: Make sure the list of filteredWords is not caps sensitive and parses input string before any command is checked
any string compared to filteredWords should be converted to lowercase first

Further, the routeMessage function should return when an offending word is found.

geisterfurz007:
Fixes #149 
Fixes #165 
Fixes #169 

Properly fetching members with their presence also requires the Privileged Presence Intent to be activated in the Discord Application under https://discord.com/developers

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@discordjs/collection": {
-      "version": "0.1.5",
-      "resolved": "https://registry.npmjs.org/@discordjs/collection/-/collection-0.1.5.tgz",
-      "integrity": "sha512-CU1q0UXQUpFNzNB7gufgoisDHP7n+T3tkqTsp3MNUkVJ5+hS3BCvME8uCXAUFlz+6T2FbTCu75A+yQ7HMKqRKw=="
+      "version": "0.1.6",
+      "resolved": "https://registry.npmjs.org/@discordjs/collection/-/collection-0.1.6.tgz",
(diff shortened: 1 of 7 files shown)


findtravishere (approved):
return statement lgtm 😄

dimapanf (approved):

PR #178, opened 2020-11-13
Feat/trends

shagunmistry:


src/programs/TopicManager.ts
```diff
@@ -35,6 +44,55 @@ export default async function TopicManager(message: Message) {
...
+          lastUsed: new Date(),
+        });
+
+        try {
+          await topicRepo
+            .createQueryBuilder()
+            .insert()
+            .into(TopicManagerEntity)
+            .values([newTopic])
+            .execute()
```
geisterfurz007, line 82: I *think* this can be shortened to 
```suggestion
            .save(newTopic)
```

src/programs/TopicManager.ts
```diff
@@ -8,7 +14,10 @@ const MAKEUP_CHALLENGE_PICTURE_URL =
 const INKTOBER_IMAGE_URL =
   "https://media.discordapp.net/attachments/689589403189772291/761243856946987078/2020promptlist.png";
 
-export default async function TopicManager(message: Message) {
+export default async function TopicManager(
+  message: Message,
+  addition?: boolean
+) {
```
geisterfurz007, line 20: I think it would be cleaner to have a seperate function for trend setting. That way we also have mod-only <-> all member functionality split and it's easier to follow the code if I don't have to recall some additional parameters that switch to something completely different in the same function (one thing responds with the current topic, the other one does database manipulation).

src/entities/TopicManager.entity.ts
```diff
@@ -0,0 +1,20 @@
...
+  id: string;
+
+  @Column()
+  topic: string;
+
+  @Column()
+  channel: string;
+
+  @Column({ default: new Date(), name: "last_used" })
+  lastUsed: Date;
```
geisterfurz007, line 15: This is never updated; maybe `created` would be a better fit?

src/entities/TopicManager.entity.ts
```diff
@@ -0,0 +1,20 @@
+import { Entity, PrimaryGeneratedColumn, Column, getConnection } from "typeorm";
+
+@Entity()
+export class TopicManagerEntity {
```
geisterfurz007, line 4: Just `Topic` should suffice, I think (same for the filename)

src/entities/TopicManager.entity.ts
shagunmistry: done, thanks

src/programs/TopicManager.ts
shagunmistry: updated. I was trying to make it fit into one method but yeah this'd be cleaner.

geisterfurz007 (approved):
LGTM!

Issue #145, opened 2020-10-07
Prohibit !voice host ing someone without Yes Theory role


geisterfurz007:
The voice command can only be used by users with the Yes Theory role. Currently `!voice host` allows an room owner to give their access to someone without that role, rendering the room frozen because the new owner cannot interact with the room.

The bot uses this code to check if the new host has the Yes Theory role:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/0d6cd14e87aec5f493eca7976c9503e1197e2cc4/src/programs/VoiceOnDemand.ts#L336-L342

It contains a subtle but bad mistake: `member` should be the mentioned user.

## Implementation

Change
https://github.com/Yes-Theory-Fam/yesbot-ts/blob/0d6cd14e87aec5f493eca7976c9503e1197e2cc4/src/programs/VoiceOnDemand.ts#L313
```diff
- const mentionedUser = message.mentions.users.first();
+ const mentionedMember = message.mentions.members.first();
```
and rename the variable in other places as well (there actually is a difference between a Member (which contains server related information like roles) and a user (which is a "plain" user of Discord, not specific to a certain server)).

In the condition mentioned above, replace member with `mentionedMember` to check the roles of the correct person.

Lastly, you need to fix a call further down:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/0d6cd14e87aec5f493eca7976c9503e1197e2cc4/src/programs/VoiceOnDemand.ts#L346

As the function expects a User object, not a GuildMember, you need the following change:

```diff
-  await transferOwnership(repo, mapping, mentionedUser, memberVoiceChannel);
+ await transferOwnership(repo, mapping, mentionedMember.user, memberVoiceChannel);
```

geisterfurz007:
This was fixed in https://github.com/Yes-Theory-Fam/yesbot-ts/commit/ea7fa734c1a30b33c3be17fd05dcfe8ce1f46d12

geisterfurz007:
This was not fixed in https://github.com/Yes-Theory-Fam/yesbot-ts/commit/ea7fa734c1a30b33c3be17fd05dcfe8ce1f46d12!

geisterfurz007:
Closed in fdb687cf3094ef90ffb6a54b035e68201893ef74

Issue #37, opened 2020-05-06
at-group pings people no longer on the server and/or deleted accounts


sklirg:
When people delete their accounts from Discord or leave the server, we do not clean up their information in our separate database. This means that we keep pinging non-existing user accounts. While it is not a problem to do so, it looks weird when people are mentioned and some of the mentions have their names show up and some have `<@!snowflake>`.

Suggestion:
Listen for guild member remove gateway event[0], and do cleanup when that happens. Verify that this is sent for a deleted user as well.

Other suggestions welcome.

0: https://discord.com/developers/docs/topics/gateway#guild-member-remove

jamie-legg:
This is definitely something that requires development. The entire flow around `guildMemberDidLeave` could use revamping to maybe even backup roles, nicknames, groups, and then have them restored once that user ID rejoins. I'll leave this up and create a task in the backlog.

geisterfurz007:
This was addressed by #80 but still needs some cleanup to remove people from the database that are not on the server anymore

geisterfurz007:
The method that removes members from user groups is flawed; here is the code:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/cb93476495e8a0196d53902ab897858efc4b3602/src/events/MemberLeave.ts#L59-L92

You can see that starting line 67, the groups are iterated under the assumption that they are group names which they are not. Because of that, the `findOne` call in line 68 does not return anything and the user who left is not removed from the group.

## Implementation

This can be fixed by removing lines 68-78 and changing the name of the parameter in the `forEach` function to `groups`. You can also remove the `async` keyword since it's not used anymore.

geisterfurz007:
Closed in 87b1e2cc7edff3414dbac2f74efed2a7a060f448

PR #215, opened 2021-04-27
fixed bug #145 and bug #37

hasanlatch007:
For the issue on #37, I have tested it but still, need to be looked at quickly to make sure it's all pretty :)


geisterfurz007:
Looks good to me, thanks!

I picked the two fix commits onto master as 87b1e2cc7edff3414dbac2f74efed2a7a060f448 and fdb687cf3094ef90ffb6a54b035e68201893ef74; I was wrong with my suggestion of deleting the package-lock.json. I got confused because I use a different package manager at work and in several other projects which produces a different lock file :)
Closing this PR since the important commits are on master now 🎉

Thanks again for your help!

PR #217, opened 2021-04-28
Change Joystick with Dice

therealMRBK:
The joystick had a weird icon when room gets created and it's hard to see what it is. Looked like a dildo to me.
A dice should improve visibility and understanding, especially in discord dark mode.


geisterfurz007 (approved):
Just checked and the joystick does look awful indeed! LGTM, thanks!

Issue #214, opened 2021-04-23
Voice channel claim can be claimed by member not in voice


hasanlatch007:
By spamming the reaction button the bot will give the spamming member the channel. We had multiple users spam it and in the end, one user did get it even though he isn't in the voice. 

Recorded if you're interested in the video my discord Duck's friend#6969 :)

hasanlatch007:
After investigating more into it, the issue is if a person not in the channel reacts and a member in the voice channel reacts to claiming the bot will take the first member who reacted and set them host

findtravishere:
I spent a while looking into it, it seems to already be written in to check that the user has to be in the voice channel on line 498 in VoiceOnDemand.ts 

```ts
  const filter = (reaction: MessageReaction, user: User) =>
    reaction.emoji.name === claimEmoji && getMemberIds().includes(user.id);
```

Not too well versed with how discord handles the user id of the person who reacts so I couldn't seem to rectify it other than encountering more bugs while testing 🤔

hasanlatch007:
Lets say User A is in a channel where the host left the channel. 
User B is not in said channel.
Yesbot sends the message asking if someone wants to claim, User B thinks he's funny and tries to claim it. Now because of the code travis send, yes the channel isn't claimed by User B
Now User A clicks on the reaction after User B clicked on it. And the channel is given to User B
I suspect this is the cause of the .first on this line

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/9a3a4fbed7836f7c956cb2e3ce1caa5fc9f012c5/src/programs/VoiceOnDemand.ts#L529

The value returned is the User B who claimed it and it bypass the filter.
I think a good fix for this would be using arrays? if the first value isn't a member in the voice channel go to second value?

geisterfurz007:
Yeah, that line is very likely to be the culprit. While the collector itself uses the filter function, `claim.users` will still contain all users who reacted, regardless of whether they are in the room or not. I guess replacing the filter checking for the bot with `.filter((user) => getMemberIds().includes(user.id)` should do the trick.

hasanlatch007:
Im on it :)

geisterfurz007:
Fixed with https://github.com/Yes-Theory-Fam/yesbot-ts/commit/4e9ad659de06c72537bbed207d1ddc96126cc60f

PR #218, opened 2021-04-28
fixed bug #214

hasanlatch007:
Added the .first() at the end because it returned a collective string and that was big red everywhere
This was only tested with 3 accounts A and B in the voice channel and C not in the voice channel trying to claim
When A (Host left) C clicked on the claim message but once pressed by B it was given to B :)
Quick warning: With such a small testing team which is me and my other accounts that I call friends. I have no idea how this will work if there are multiple in the voice channel. 
But going off logic here seems okay?


geisterfurz007:
LGTM! Thanks for the help finding a reproduction of this and implementing the fix :)

Issue #219, opened 2021-04-29
!group search {random group} only makes the bot reply with the user who typed the command  and !group update called !group changeCooldown at the same time.


hasanlatch007:
I have already written a fix for this but wanted to still create an issue to log this better with a proper description
When a user writes a random group in the command !group search (anything that isn't a group) the bot just replies with the user's tag

I believe the cause was in the function itself and the problem was in this line
https://github.com/Yes-Theory-Fam/yesbot-ts/blob/4e9ad659de06c72537bbed207d1ddc96126cc60f/src/programs/GroupManager.ts#L377-L378

Because the .findOne didn't return any value it kept running the code but with a group value of null which doesn't print the lists of groups because of the code before it basically the whole function

I fixed this by adding certain if conditions, I think it could've done way better than how I did it but it works right?

The second issue was a missing `break;` at

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/4e9ad659de06c72537bbed207d1ddc96126cc60f/src/programs/GroupManager.ts#L106-L114

Which just ran the changeCooldown case also fixed

Basically the new additions all it does is:

if a user searches a nonexistant group he gets a reply saying the group doesn't exist
If a user just types !voice search it replies with the lists of groups
If a user types !voice search (an existing group) it just does what its suppose to do 

and added the break; pretty straightforward.

geisterfurz007:
Closed with 7d0f36c4595cb744c70eeb5d98277ec660311e96

PR #221, opened 2021-04-29
Here is the updated typeorm with the new ilike features added bot runs smoothly and also enhancements for issue #219

hasanlatch007:
Ran the code using the 4 commands worked like a charm
The new typeORM works because well groupmanager uses iLike and it didn't break :D
I added the PNG format but this hasn't been tested!!! As @Someone doesn't work for me I guess
Did a few enhancements for !group I'm sure this could've been written better but it works!

Sorry for spamming the logs but GitHub kept fucking around so I had to delete everything multiple times and rewrite everything :)


src/programs/GroupManager.ts
```diff
@@ -295,97 +296,104 @@ const searchGroup = async (
...
-
-  const copy = (
-    await groupRepository.find({
-      where: {
-        name: ILike(`%${requestedGroupName}%`),
-      },
-      relations: ["members"],
-    })
-  ).sort(byMemberCount);
+  if (requestedGroupName === "" || group) {
```
geisterfurz007, line 305: This causes only exact matches to result in output and requires two database queries. We are already fetching matching groups a few lines down below (the stuff starting with `const copy = (`) so checking whether the length of that is 0 and responding with something like "No matching groups were found" is the better option here.

Also to avoid adding nesting (here in the diff you can see many lines being added because they had 2 spaces added before them), you can use an early return statement like this:

```ts
const copy = (
  await groupRepository.find({
    where: {
      name: ILike(`%${requestedGroupName}%`),
    },
    relations: ["members"],
  })
).sort(byMemberCount);

if (copy.length === 0) {
  await message.reply("No matching groups were found");
  return;
}

// At this point it is guaranteed that we found at least one group so the code from here on can remain as before
const pageAmount = Math.ceil(copy.length / groupsPerPage);
```

geisterfurz007 (changes requested):

hasanlatch007:
New commit: Removed the two if and replaced it with the `if (copy.lengh=== 0)` and fixed the extra spaces

geisterfurz007 (approved):
LGTM!

Issue #222, opened 2021-04-30
Fix several "Cannot read property x of undefined" issues in Spyfall


geisterfurz007:
This is pretty much all I have, gotta find some time to find and fix this as it leads to game sessions not being closed properly.

![grafik](https://user-images.githubusercontent.com/26303198/116736925-d8d4c500-a9f0-11eb-8519-101ddc5d6d44.png)

geisterfurz007:
Closed with 09c02cb4df4b2661039d3a766b6373e8b8287aba

Issue #223, opened 2021-04-30
Don't force add game master if they are in a session already


geisterfurz007:
There are several bugs in the GameHub at the moment all around a similar topic:

 - [ ] The game master is added to the game despite already being in a session (I think)
 - [ ] An empty session is still started (related to the item below but a slightly seperate case since there can be no minPlayers in the game options)
 - [ ] The minPlayers aren't respected

PR #224, opened 2021-05-02
Fixes for issue #222 and issue #223

hasanlatch007:
The reason for this bug was because the bot is not able to send to the player a dm if they have it disabled in their privacy settings, it generated a bunch of errors in the log making a non-ending session.

Testing:
4 accounts to launch the game. 
- First Test: One account had their privacy disabled the bot send a message in #bot-games text channel saying the game couldn't start because of @{PLAYER WHO HAS "ALLOW DIRECT MESSAGES FROM SERVER MEMBERS" DISABLED} and wouldn't launch the game. Trying to open another game session immediately generated no bugs and also didn't allow the game to start.
- Second Test: All users had the right settings game launched with no issue, all commands work normally, the game runs with configurations being changeable. Tbh idk what I missed but doesn't feel like did...

Implementations:
Added a new configuration: redquiredDms (Allows you to run the validates function).
ValidateDms is a function that sends a text message to all players in the game to make sure they can receive a message
Updated Spyfall to use the new configuration.
Updated GameSession to use the new configuration.
Non imported I just added the `;` in the GroupManager because it pissed me off that I forgot about it.

### DIFF
--- a/src/games/GameHub.ts
+++ b/src/games/GameHub.ts
@@ -200,6 +200,43 @@ export default class GameHub {
     }
   }
 
+  async validateDms(players: GuildMember[]) {
+    const safeDm = async (
+      user: GuildMember,
+      message: string
+    ): Promise<boolean> => {
+      try {
+        const dm = await user.createDM();
+        await dm.send(message);
+        return true;
+      } catch (e) {
+        return false;
+      }
+    };
+    // Map all players to a Promise<{ player, sendSuccess: bool }>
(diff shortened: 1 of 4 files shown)


geisterfurz007:
LGTM! Note: This only fixes #222. #223 is somewhat unrelated and came up as a result of #222 because players tried to create a new game. I updated #223 with more information on what I am fairly sure is the cause.

PR #225, opened 2021-05-03
fix: issue #223

hasanlatch007:
Testing:

Tested users currently in a game and if they can be invited by user C, the double-check for issue #223 basically
Tested users cannot be invited to a new game if on the configuration page

Implementation:

In file: GameHub.ts, added the if to check if cleanPlayers is bigger or equal to the minimum of players required before starting the game.
In these files: GameHub.ts, GameSession.ts, and Spyfall.ts. Added the new property called Configuration. Basically split the patch config to avoid users being invited to a new game while on the configuration page of the game.

### DIFF
--- a/src/games/GameHub.ts
+++ b/src/games/GameHub.ts
@@ -273,7 +273,7 @@ export default class GameHub {
       throw e;
     }
 
-    const minPlayers = clazz.config;
+    const minPlayers = clazz.config.minPlayers;
     const cleanedPlayers = await this.cleanPlayers(players);
     const permissions = this.getChannelPermissions(cleanedPlayers);
     const channel = await this.createChannel(config, permissions);
@@ -282,7 +282,7 @@ export default class GameHub {
       .map((member) => `<@${member.user.id}>`)
       .join(" ");
 
-    if (cleanedPlayers.length < clazz.config.minPlayers) {
+    if (cleanedPlayers.length < minPlayers) {
       await channel.send(
         `Not enough players! You need at least ${minPlayers} people.`
       );
--- a/src/games/GameHub.ts
+++ b/src/games/GameHub.ts
@@ -253,7 +253,7 @@ export default class GameHub {
         players.push(this.guild.member(leaderId));
       }
 
-      const { minPlayers, maxPlayers } = clazz.config;  
+      const { minPlayers, maxPlayers } = clazz.config;
       if (minPlayers && players.length < minPlayers) {
         // noinspection ExceptionCaughtLocallyJS
         throw new Error(
@@ -281,7 +281,7 @@ export default class GameHub {
 
     if (cleanedPlayers.length < minPlayers) {
       throw new Error(
-        `Not enough players! You need at least ${minPlayers} people.`
+        `Not enough players! You need at least ${minPlayers} people. Please make sure no players are in an ongoing game.`
       );
     }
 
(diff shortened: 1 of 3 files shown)


src/games/GameHub.ts
```diff
@@ -274,51 +274,54 @@ export default class GameHub {
     }
 
     const cleanedPlayers = await this.cleanPlayers(players);
-    const playerPing = cleanedPlayers
-      .map((member) => `<@${member.user.id}>`)
-      .join(" ");
+    if (cleanedPlayers.length >= clazz.config.minPlayers) {
```
geisterfurz007, line 277: Invert the if to reduce nesting (so `if (cleanedPlayers.length < clazz.config.minPlayers) { /* send message and return */ }`)

src/games/GameSession.ts
```diff
@@ -93,8 +93,7 @@ export abstract class GameSession<T extends SessionConfig> {
       .map((user) => this.hub.guild.members.resolve(user.id));
   }
 
-  public async patchConfig(
-    partial: Partial<T>,
+  public async Configuration(
```
geisterfurz007, line 96: Function names should start with a lowercase letter. This function doesn't have to be async. Also just configuration is quite vague, how about
```suggestion
  public setBaseConfiguration(
```
The return type would then just be `void`.

geisterfurz007 (changes requested):

src/games/GameHub.ts
hasanlatch007: Okay, what should I set the send message as?

geisterfurz007: With the same one that we use already a little further up in the code:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/09c02cb4df4b2661039d3a766b6373e8b8287aba/src/games/GameHub.ts#L260

hasanlatch007: Oki used the right one sending it your way!

hasanlatch007:
Also on the GameHub.ts I changed the variables to the top of the if at line 285 to make it look better and neattier

src/games/GameHub.ts
```diff
@@ -273,14 +273,21 @@ export default class GameHub {
       throw e;
     }
 
+    const minPlayers = clazz.config;
```
geisterfurz007, line 276: This is incorrect. Either: `const minPlayers = clazz.config.minPlayers;` or `const { minPlayers } = clazz.config;` would work however.

src/games/GameHub.ts
```diff
@@ -273,14 +273,21 @@ export default class GameHub {
...
+    const channel = await this.createChannel(config, permissions);
+    let maybeVoiceChannel;
     const playerPing = cleanedPlayers
       .map((member) => `<@${member.user.id}>`)
       .join(" ");
 
-    const permissions = this.getChannelPermissions(cleanedPlayers);
-    const channel = await this.createChannel(config, permissions);
-    let maybeVoiceChannel;
+    if (cleanedPlayers.length < clazz.config.minPlayers) {
```
geisterfurz007, line 285: Since you extracted minPlayers to a variable, you can use it here as well:
```suggestion
    if (cleanedPlayers.length < minPlayers) {
```

geisterfurz007 (changes requested):

hasanlatch007:
Should be fixed now!

hasanlatch007:
Tested it, it does delete the channel after 60 seconds and now the message is sent in #bot-games

geisterfurz007 (approved):

Issue #208, opened 2021-03-04
Non-owners cannot join voicechat when it gets created


dimapanfilov:
After owner joins, the VC stays locked. If the owner leaves and rejoins then that fixes the issue

Frequency: More often lately 

Maybe more discord inconsistency 🤔 ?

geisterfurz007:
Something that might cause the issue is a race condition between bot and user.  
If the owner joins before the database entry is saved, the voice state handler will just return because no matching room was found. That would also explain why it worked when the owner leaves and rejoins because then the database entry exists and the bot functions as expected.

This can be checked by spam clicking on the room as it's being created to be faster than the bot and trying to get in before the message "Your room [...] was created." is sent. In that case the room is likely to not get unlocked.

If that's the case, the fix is surprisingly straightforward: Move creation of the database entry further up; ideally between creation of the channel and setting the permissions that allow the owner to join. That way the owner will only be able to join once the entity is saved to the db.

geisterfurz007:
This rarely happens in general and hasn't happened since a month so I am closing this, hoping the issue is indeed fixed in https://github.com/Yes-Theory-Fam/yesbot-ts/commit/d05278360ba05eed7a4097ad85063cac2f550813

PR #227, opened 2021-05-08
Fix issue: #208

hasanlatch007:
Tried clicking as fast as possible for this couldn't replicate it with the new code. Not sure if I couldn't replicate it properly or It's fixed would recommend testing it a bit more!

### DIFF
--- a/src/games/GameHub.ts
+++ b/src/games/GameHub.ts
@@ -285,15 +285,13 @@ export default class GameHub {
       );
     }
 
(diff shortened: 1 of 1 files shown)


hasanlatch007:
Also ignore the bunch of commits github is being a lil bich

geisterfurz007:
LGTM! It certainly won't break anything and I couldn't repro either with spam clicking my poor mouse so we'll give this a shot. Don't worry about the commits, they will end up as one on master thanks to git magic :)

Issue #205, opened 2021-02-21
Implement a way to add new deadchat questions


geisterfurz007:
Currently the only way to add more !deadchat questions is to insert them directly into the database. This should be possible in another way.

geisterfurz007:
This is currently not (cleanly) possible due to https://github.com/prisma/prisma/issues/2506

Given this feature isn't used often and adding deadchat questions manually isn't much of a problem, I would like to wait for a native Prisma solution for now.

Issue #189, opened 2020-12-25
Replace TypeORM Repositories with BaseEntity


geisterfurz007:
Instead of creating repositories for entities, we can make the entities `extend BaseEntity` (from typeorm) and then use `await Birthday.find(userId)` for example instead of
```
const repo = await BirthdayRepository();
await repo.find(userId);
```
which should save a good bunch of code and make things more readable. All other functions exposed by the repository (like update and delete) are also available on the class when it extends `BaseEntity`.

## Partially completed

This issue was partially completed with #228, remaining entities are listed here:

- [ ] `ChannelToggle`
- [ ] `DailyChallenge`
- [ ] `Message`
- [ ] `ReactionRole`
- [ ] `Topic`
- [ ] `VoiceOnDemandMapping`

`BuddyProjectEntry` is **explicitly *not* part of this issue**, since the entirety of the buddy project related code will be rewritten in the future and is currently unused so there is nothing to gain.

## Implementation

All database entities are placed in `src/entities`. Most of them currently do not inherit from `BaseEntity` and instead export a `EntityNameRepository` function that creates a new repository to modifiy the respective entities. To change the entities to use `BaseEntity` following these steps should work:
 - Import `BaseEntity from "typeorm";`
 - Make the entity class `extend BaseEntity` (if it doesn't already)
 - Remove the exported function that creates a repository
 - Find all occurences of the function being imported and used, remove them and replace all things that are like `repo.update` for example with `EntityName.update`.
 - `EntityName.save` can be simplified as `entity.save`

geisterfurz007:
Closing since the bot is moving from TypeORM to Prisma

PR #228, opened 2021-05-08
Convert database entities from TypeORM Repositories to extending BaseEntity

sklirg:
Resolves _parts of_ #189.

I think the rest can follow suit, but I am not sure how to test that it all works. These changes worked out the box based on the implementation description you gave in #189, so I think that will be the case for the remaining entities too.

Merging this unblocks #205.


geisterfurz007:
LGTM! Awesome, thanks! I will keep #205 blocked for now still since the plan is to achieve a common solution for Deadchat and DailyChallenge and the latter is still repo based but this brought us closer one step for sure!

Issue #235, opened 2021-05-21
Add Tests to the bot


adrian-goe:
Currently, there are no automatic tests.

To make sure that working code continues to work, tests should be added.

- [ ] add test framework (jest)
- [ ] write test for https://github.com/Yes-Theory-Fam/yesbot-ts/pull/242 after it is merged.

This will be the foundation, so all further refactoring can write test directly

sklirg:
This should be pretty easy to convert to a testing framework when one is decided and set up: https://github.com/Yes-Theory-Fam/yesbot-ts/blob/master/src/programs/BirthdayManager.spec.ts

adrian-goe:
@sklirg I assume so, but since we refactor more or less the whole code anyway, the test will rather not fit the new structure of the code itself.
But we will see how the code looks then.
Thanks for the hint 😊

PR #231, opened 2021-05-21
Move to Prisma ORM

geisterfurz007:
This PR (hopefully) contains all changes required to move the bot from TypeORM to Prisma using [`@yes-theory-fam/database`](https://github.com/Yes-Theory-Fam/database).

It also removes all remains of the Buddy Project and fixes missing `await` in many, many, many, many places.  
Since the npm package is private, the CI will fail trying to download it currently but all files are properly formatted (pinky promise :) ).

PR #237, opened 2021-05-21
setup jest with first basic jest tests

adrian-goe:
- setup jest for https://github.com/Yes-Theory-Fam/yesbot-ts/issues/235
- add first tests


adrian-goe:
https://github.com/Yes-Theory-Fam/yesbot-ts/issues/235

geisterfurz007:
After the additions from #231, the tests don't run anymore since jest cannot resolve type-graphql anymore. I just published `@yes-theory-fam/database@0.4.4` that allows direct referencing of the client through the import `@yes-theory-fam/database/client`. Replacing all `@yes-theory-fam/database` imports with that makes jest run as expected.

__tests__/common/CustomMethods.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+      id: 312297787901607937,
+    },
+    mentions: {
+      users: mentionUsers,
+    },
+    channel: { send: jest.fn() },
+    reply: jest.fn(),
+    react: jest.fn(),
+    cleanContent: "cleanContent",
+  } as unknown) as Message;
```
geisterfurz007, line 27: This might be worth extracting to a function with a few arguments for things like author, content and cleanContent, since it's already duplicated a little further below.

__tests__/common/CustomMethods.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+describe("test custom methods", () => {
+  const mockDiscord = new MockDiscord();
+
+  const mentionUsers = new Collection<Snowflake, User>();
+  mentionUsers.set("Discord.SnowflakeUtil.generate(new Date()).toString()", mockDiscord.getUser());
+
+  const message: Message = ({
+    author: {
+      id: 312297787901607937,
```
geisterfurz007, line 18: It doesn't matter much but I think this could be a more generic ID

jest.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
```
geisterfurz007, line 3: `__tests__/mocks/index.ts` was picked up as a testfile and jest complained that there was no test in it.
`modulePathIgnorePatterns: ["<rootDir>/__tests__/mocks"],` in here excludes the mocks.

geisterfurz007 (changes requested):
Other things:

 - Files are not formatted with prettier
 - The test never terminates and appears to hang infinitely, not sure about the cause here

__tests__/common/CustomMethods.spec.ts
adrian-goe: No, I wouldn't move this to a function, because the mocks are built specifically for the tests. That would mean a lot more effort.

adrian-goe:
> * The test never terminates and appears to hang infinitely, not sure about the cause here
I'm trying to find a solution for that. But currently I dont find where the `DiscordjsError: Request to use token, but token was unavailable to the client.` is thrown, because the test is successful. But I will look further into this.

adrian-goe:
did a new PR for the new structure (https://github.com/Yes-Theory-Fam/yesbot-ts/pull/278) so this will be closed.

PR #242, opened 2021-05-27
chore: reference implementation for a generic event handler

geisterfurz007:
This PR contains a possible start of a rewrite of YesBot's event handling. The plan is to load all event handlers (which implement a central class) to a dedicated handler class which then takes care of picking the correct handlers and instantiating them if needed.

The loading and configuration of handlers is done using a custom decorator that adds the configuration to Rambo's handler tree.

I think Rambo is a monstrosity and completely overengineered (while I am not even sure it convers all usecases :') ) but maybe we can build on top of it.

Why Rambo? I jokingly named it somewhere during development and I didn't find a better name for it :') Suggestions welcome although I would also not have a problem with leaving the name as it is.

TODOs:
 - [ ] Implement permission checks
 - [x] Implement handling for DM only commands
 - [ ] Extend the handler for other events
   - [x] ReactionAdd
   - [x] ReactionRemove
   - [ ] GuildMemberUpdate
   - [ ] MemberJoin
   - [ ] MemberLeave
   - [ ] Ready
   - [ ] VoiceStateUpdate
 - [ ] Port existing programs over (can be done incrementally, keeping the current event handlers intact and removing programs from the imperative handler once moved to the decorator based one)
 - [ ] Remove DecoratorTest


adrian-goe:
_To remain transparent, here is my critique from the DM_

- I would write the interfaces and the enum in a separate file.
- as well as the command handler
- types can be written in a common file? But maybe also in a separate file.
- Enum values are written in best practice full uppercase.
- the triggers as enum with string values? Then they are all collected in one place. Maybe a little clearer.

adrian-goe:
As this is a major refactoring of the existing codebase, it could be a good time to introduce a naming convention.
Mixed naming like `index.ts` and `CustomMethods.ts` is maybe not a good way.

My proposal would be something similar like in angular. 
This means:
Filenames are in kebab-case `custom-methods.ts`
Classnames are in PascalCase `export default class CustomMethods {`
Variables and Functions are in camelCase `const isTrue;` `getIsTrue();`

Feel free to discuss this 😊

geisterfurz007:
- Created new folder for rambo under `src/rambo` (I don't know if I will get rid of that name on my own, help 😄)
- Interfaces, enums and the abstract handler class were moved under `src/rambo/types`:
  - `base.ts` contains the enum and several base types
  - `handler.ts` the class and a type definition for the class constructor
  - `hioc.ts` contains type definitions used to create the tree of handlers within rambo on init
- Types and functions for events are in `src/rambo/events`. It currently contains implementations and types for new messages, as well as added and removed reactions. Each file (aside `index.ts`) contains the following definitions for the respective event(s):
  - Options for the decorator
  - Type to define the handler function's parameters
  - A function adding a handler to the StringIndexedHIOCTree of the event
  - A function to extract the information from an event that Rambo needs to determine the correct handler(s)

I don't think we should keep triggers in one place. I'd rather add a script somewhere in the future that uses the decorators to auto-generate documentation for all sorts of events including triggers for messages. In pretty much all cases they will be one-time use anyway and would just mean another file to touch when creating a new command. One of the main goals I had with this refactor is making the bot more modular so that each file represents a functionality standalone which can be removed without having to edit any other files.

- Renamed all files to kebab-case
- Converted constructors-turned-function to arrow functions
- Converted all `export default async function PascalCase` to `const camelCase = async () =>`
- I left some "compile-time" constants in Snake/Caps and left the `Command` decorator named this way since I am used to decorators being in PascalCase (I can't help it :') )

src/programs/someone.ts
```diff
@@ -7,7 +7,7 @@ import prisma from "../prisma";
 const QUESTION_LINK: string =
   "https://spreadsheets.google.com/feeds/cells/1eve4McRxECmH4dLWLJvHLr9fErBWcCGiH94ihBNzK_s/1/public/full?alt=json";
 
-async function Someone(message: Message) {
+const someone = async (message: Message) => {
```
adrian-goe, line 10: Is there a better name? Without Context, I would not know what happens here

src/rambo/types/base.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const enum DiscordEvent {
+  MESSAGE = "MESSAGE",
+  REACTION_ADD = "REACTION_ADD",
+  REACTION_REMOVE = "REACTION_REMOVE",
+  // GUILD_MEMBER_UPDATE = "GUILD_MEMBER_UPDATE",
+  // MEMBER_JOIN = "MEMBER_JOIN",
+  // MEMBER_LEAVE = "MEMBER_LEAVE",
+  // READY = "READY",
+  // VOICE_STATE_UPDATE = "VOICE_STATE_UPDATE",
```
adrian-goe, line 14: Could be deleted since it is already in git history

adrian-goe:
> * Created new folder for rambo under `src/rambo` (I don't know if I will get rid of that name on my own, help 😄)

Would prefer a meaningful name buts its okay 😄

> I don't think we should keep triggers in one place. I'd rather add a script somewhere in the future that uses the decorators to auto-generate documentation for all sorts of events including triggers for messages. In pretty much all cases they will be one-time use anyway and would just mean another file to touch when creating a new command. One of the main goals I had with this refactor is making the bot more modular so that each file represents a standalone functionality which can be removed without having to edit any other files.

See it the same way. For me, it would be enough, when the files are in the dame directory.

> * I left some "compile-time" constants in Snake/Caps and left the `Command` decorator named this way since I am used to decorators being in PascalCase (I can't help it :') )

This is fine. I haven't seen a decorator other than PascalCase

src/programs/someone.ts
geisterfurz007: `atSomeone` or `someoneTag`?

src/events/guild-member-update.ts
```diff
@@ -13,56 +12,50 @@ import { hasRole } from "../common/moderator";
...
-      member.roles.cache.some(({ name }) => {
-        return regionCountries.some((country) =>
-          name.includes(`${country}! (`)
-        );
-      });
+const guildMemberUpdate = async (
+  oldMember: GuildMember | PartialGuildMember,
+  newMember: GuildMember | PartialGuildMember
+) => {
+  const regionCountries = ["Australia", "Canada", "the UK", "the USA"];
```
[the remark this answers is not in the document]
geisterfurz007: I think I thought I needed it but that's indeed not the case, good catch!

src/events/message.ts
```diff
@@ -0,0 +1,239 @@
...
+
+const message = async (msg: Message) => {
+  if (msg.channel.type === "dm" && !msg.author.bot) {
+    await routeDm(msg);
+  } else {
+    await routeMessage(msg);
+  }
+};
+
+const routeMessage = async (message: Message) => {
```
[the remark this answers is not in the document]
geisterfurz007: I hope to get rid of most of that in this one (or follow-ups) replacing the switch and ifs with the decorator and trusty rambo.

adrian-goe (commented):
LGTM

src/events/message.ts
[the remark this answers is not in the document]
adrian-goe: oke 👍🏻

src/programs/someone.ts
adrian-goe: I guess somoneTag fits better

src/event-distribution/events/message.ts
```diff
@@ -62,5 +50,6 @@ export const extractMessageInfo: ExtractInfoForEventFunction<DiscordEvent.MESSAG
     return {
       handlerKeys: [channelIdentifier, message.content.split(" ")[0]],
       user: message.author,
+      isDm: message.channel.type === "dm",
```
adrian-goe, line 53:
```suggestion
      isDirectMessage: message.channel.type === "dm",
```

src/event-distribution/events/reactions.ts
```diff
@@ -57,5 +57,9 @@ export const extractReactionInfo: ExtractInfoForEventFunction<
 
   const channel = reaction.message.channel;
   const channelIdentifier = getChannelIdentifier(channel);
-  return { handlerKeys: [channelIdentifier, reaction.emoji.name], user };
+  return {
+    handlerKeys: [channelIdentifier, reaction.emoji.name],
+    user,
+    isDm: reaction.message.channel.type === "dm",
```
adrian-goe, line 63:
```suggestion
    isDirectMessage: reaction.message.channel.type === "dm",
```

src/event-distribution/types/base.ts
```diff
@@ -9,12 +9,19 @@ export const enum DiscordEvent {
   REACTION_REMOVE = "REACTION_REMOVE",
 }
 
+export const enum MessageLocation {
+  SERVER = "SERVER",
+  DM = "DM",
```
adrian-goe, line 14:
```suggestion
  DIRECT_MESSAGE = "DM",
```

src/event-distribution/types/base.ts
```diff
@@ -30,6 +37,7 @@ export type HandlerFunctionFor<
 export interface HandlerInfo {
   handlerKeys: string[];
   user: User;
+  isDm: boolean;
```
adrian-goe, line 40:
```suggestion
  isDirectMessage: boolean;
```

src/events/ready.ts
```diff
@@ -29,14 +29,14 @@ const ready = async (bot: Client) => {
...
-    });
-    readyMessage.edit(readyMessageString("Members fetched, fully ready!"));
+    // const readyMessage = await outputChannel?.send(
+    //   readyMessageString("Fetching members.")
+    // );
+    //
+    // await guild.members.fetch({
+    //   withPresences: true,
+    // });
+    // readyMessage.edit(readyMessageString("Members fetched, fully ready!"));
```
adrian-goe, line 39:
```suggestion
```

src/event-distribution/README.md
```diff
@@ -34,10 +34,10 @@ interface BaseOptions {
...
   location?: MessageLocation; // A location enum setting where events are accepted.
   // Available values are:
-  //  - MessageLocation.SERVER (only call handler for events on the server)
-  //  - MessageLocation.DIRECT_MESSAGE (only call handler for events in DMs) and 
-  //  - MessageLocation.ANYWHERE (call handler regardless of the origin of the event).
-  // The default is MessageLocation.SERVER if channelNames is non-empty, MessageLocation.ANYWHERE otherwise.
+  //  - EventLocation.SERVER (only call handler for events on the server)
+  //  - EventLocation.DIRECT_MESSAGE (only call handler for events in DMs) and 
+  //  - EventLocation.ANYWHERE (call handler regardless of the origin of the event).
+  // The default is EventLocation.SERVER if channelNames is non-empty, EventLocation.ANYWHERE otherwise.
```
adrian-goe: * why is the type of location still `MessageLocation`, when the enum is EventLocation?
* why do we need the comment of available values? They are in the enum and this should be self explained, with values are possible with this enum.

src/event-distribution/types/base.ts
```diff
@@ -36,7 +36,7 @@ export type HandlerFunctionFor<
 
 export interface HandlerInfo {
   handlerKeys: string[];
-  user: User;
+  member: GuildMember | null;
```
adrian-goe, line 39:
```suggestion
  member?: GuildMember;
```

adrian-goe (changes requested):

src/event-distribution/README.md
geisterfurz007: First time WebStorm has screwed me over!

It's true, we don't need them. Removed most of it, moved the default hint to the same line as the key and updated it to also mention requiredRoles as reason to default to SERVER.

Issue #244, opened 2021-06-02
!voice rooms remain available without people


geisterfurz007:
I am fairly confident that this bug is caused by a failed DB constraint. Technically you can create multiple rooms with the same user however only one is saved due to the user_id constraint.

To fix this, make sure that in all places where the ownership of a room changes, the new owner does not already own a room.

This may occur if:
 - `!voice create` is used during the time of emoji selection for a room
 - `!voice host` is used with a user who already owns a room
 - The owner of room A leaves it and a user with a room (B) joins room A and claims ownership

I ~~hope~~ think this covers all cases.

PR #246, opened 2021-06-06
fix: issue #244

hasanlatch007:
Added checks to verify if the user has a voice channel when creating a voice channel, when given host or when reacting to get ownership,

The first 2 are self-explanatory just a simple `if` using getVoiceChannel.

The mess begins at the requestOwnershipTransfer function.
I divided between 2 main `if` and one sub `if` in the first one.
The first one sees if the claimingUser has a voice channel, if he does we can divide it between 2 scenarios, he was randomly chosen or he reacted.
The sub `if` is for the first scenario: no one reacted and the random user selected has a voice channel if he does, it will randomize again without him included in the list of members, this was tested and works as functioned, I had 3 users, User A (host) and B in a voice channel while User C in his own channel. User A disconnect I wait for the transferMessage and set a timer for 60 seconds. Around 10 seconds User C joins the room, but no one has reacted. I had the code print out "1" in the console to verify it went through the scope which it did :).
If he reacted this is where our `else` kicks in, with an error message throwing out and using the same filter and choosing another user.

Lastly is a simple check if the claiminguser doesn't have a room if he doesn't everything proceeds like normal.
All of these were checked using console.log() and worked :)


hasanlatch007:
Forgot to add the delete message at line 555

hasanlatch007:
Added the fix for !voice create, because the way Discord.js works, if a user would send the message !voice create twice in a row and click separately on each emoji for both the messages, the `if (hasExisting)` check will return false even if he created a voice channel, because of Disord.js? Anyways I found the best way to fix this is by immediately deleting the second channel since it isn't registered in the database this wouldn't cause any conflicts and he would still have his first created voice channel, using `try{}` it will look for the error code "P2002" and immediately delete the channel smooth transition baby 😎 

Added more scenarios for the claim ownership, if the user has a voice channel and tries to take over another one when he's the only one in it, it will also be immediately deleted and messages him that he already owns a voice channel, avoiding any conflicts that could happen with the database.

I truly think I tackled every possibility with this and we should not suffer the voice channel not being deleted anymore.

geisterfurz007 (changes requested):

geisterfurz007:
I will check the rest tomorrow, my head isn't working properly, atm

hasanlatch007:
To be honest, I have no idea why I moved it :'), thinking of too many algorithms possible, broke my brain. Anyways removed all nesting.
Take a look at it whenever you're okay all is good!

hasanlatch007:
also typo in line 562 sorry

hasanlatch007:
Do not review the second part till I add a new commit, it is not working correctly.

hasanlatch007:
Took me a while sorry! All should be working as intended now!
Detail of the idea and why I added so much for the second part, like I said I divided all scenarios possible for reacting and breaking the bot, but to avoid nesting I also broke it :'), I also added the variable reactions, in case someone outside the vc's is fooling with the claim ownership button, this way the bot won't choose the wrong `if`

hasanlatch007:
My dumbass was commiting on master im so sorry michel

geisterfurz007 (changes requested):
There is a lot of code [here](https://github.com/Yes-Theory-Fam/yesbot-ts/pull/246/files#diff-ac5f2f17ecb2e10619499abd62e84828ba0ef65b7e1989bf9ca22a69bed51c76R549-R610) that could maybe be shortened by quite a bit using a different approach. Essentially the goal is to follow these three steps:
 - Let the members pick
 - If they did and the claimer doesn't have a room -> transfer ownership
 - If they did and the claimer does have a room -> pick at random
 - If they didn't -> pick at random
with "pick at random" implying the collection of users to be picked from is all members in the room who don't have a room yet.

So in essence you could prefilter the memberlist by requesting the following from the database:
```ts
    await prisma.voiceOnDemandMapping.findMany({
      select: { userId: true },
      where: { userId: { in: getMemberIds() } },
    })
```
That will leave you with a list of `{ userId: string }` objects where each userId is in the room that needs a new owner and has their own room too. So you can get a list of users in the channel who don't have a room by removing those from the list of all ids (I don't know if that made sense 🤔).

That should make it straightforward to implement the steps described above. With the list of "allowed" ids, you can both check if the user who claimed the room (if any) already has a room (if they are not in that list, they do) and picking at random from a list is fairly easy to do.

If the list is actually empty in case you need to pick someone at random, you can still post a message and delete (but delete after a timeout). *Technically* you could double check after a minute if one of the members in there has lost their room in the meantime but I think it's fair to just delete it. This is such an absurdly rare edge case, that the people doing it probably deserve one room less in their group ;)

Issue #142, opened 2020-10-04
Add group ping setting to only allow pings in a calm room


geisterfurz007:
Some groups are gigantic and cause a rush of messages which can be disruptive to ongoing discussions. For this reason a group should get a boolean field indicating whether a group may only be pinged when the room the ping occured in has had no message in the past 3 minutes.

## Implementation

The entity containing information about a pingable group is placed in `node_modules/yes-theory-fam/database/prisma/schema.prisma` (installed from the package [here](https://github.com/Yes-Theory-Fam/database/blob/master/prisma/schema.prisma)).  
Add a new number column called `deadTime` (or a more fitting name if you have one!) with default database value 0.  
In `src/programs/group-manager.ts` add a new command to the switch case in the exported function which allows moderators to change the value of that field for a given group. You can draw inspiration from other commands like `changeCooldown` to get an idea of what you need to do here. Be sure to add a `break;` to the switch statement when extending it.  
Finally, this is the code used to check for potential reasons not to send the group ping:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/0d6cd14e87aec5f493eca7976c9503e1197e2cc4/src/programs/GroupManager.ts#L153-L164

Here you first get the channels last message (you cannot use `message.channel.lastMessage` because it's the group ping `@group groupname` that was sent), check the time difference from `previousLastMessage.createdAt` to `Date.now()` and calculate if the difference is larger than what is set as the group's `deadTime` in minutes. If it has been less than that, use `Tools.handleUserError` to inform the user about the reason the group ping won't be sent.   
To get the last message before the command, use the same logic that !deadchat does (with a different time and maybe a few more intermediate variables 🦥):

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/cb93476495e8a0196d53902ab897858efc4b3602/src/programs/Deadchat.ts#L6-L10

PR #248, opened 2021-06-08
Feature: deadchat for groups

hasanlatch007:
This was tested and the general idea of it worked, but if I were to @group Hydro for example and the chat is not dead, I won't be allowed to ping them it works as described, but I won't be able to even ping it in a deadchat due to the groupcooldown, I didn't know if I should fix it or keep it needs more input on the idea (bugs are features).

(I'm still new to branches so I'm sorry if these commits are all over the place :'))


geisterfurz007:
The relevant issue is #142; it goes into more detail on what things would be helpful for this feature including different group timeouts for different groups. While hydrohomies certainly deserves to only be pinged in calm rooms as it's mostly used in the main chats, groups like amongus would be significantly less useful since the chats they are used in tend to be active *while* it's being played and finding a new member is often done by pinging the group.
The whole thing requires a change in the database package unfortunately which I didn't fully think through when extracting the logic to a separate repo that doesn't have automatic publishing but for testing purposes, you can edit the schema directly in `node_modules/@yes-theory-fam/database/prisma/schema.prisma` and use `npm run db:prisma db push` (I think) to update your database locally and regenerate the sources for accessing it.

adrian-goe:
closes #142

adrian-goe:
As discussed with @hasanlatch007 he is opening a new PR.

Issue #254, opened 2021-06-16
Dockerize the bot


adrian-goe:
To simplify and possibly automate the deployment, a Docker image should be created.
The image should then be deployed through the GitHub-Actions via the GitHub-ContainerRegistry.

PR #264, opened 2021-06-16
Creat dockerimage

adrian-goe:
#254 
Create the Dockerfile
net step would be to create the pipline for this


docker-compose.yml
```diff
@@ -9,3 +9,19 @@ services:
...
+      BOT_TOKEN: "ODQ1MzA0NjU5MzAzODU4MTkx.YKfBTw.h_-nhliH54bnKpsEtiCI9lz_pfE"
+      GUILD_ID: 845306344458027009
+      OUTPUT_CHANNEL_ID: 845306346810638366
+      MAP_LINK: "<>"
+      MAP_ADD_DM_USER_ID: "<>"
+      MODERATOR_ROLE_NAME: "Support"
+      ENGINEER_ROLE_NAME: "Server Engineer"
+      COORDINATOR_ROLE_NAME: "Server Coordinator"
+      BUDDY_PROJECT_MATCHING: "false"
+      PRISMA_DATABASE_URL: "postgresql://yesbot:yesbot@yesbot-ts_postgres_1:5432/yesbot?schema=public"
```
geisterfurz007, line 25: We should probably use env variables for this, then we can still configure these on the server.

tsconfig.json
```diff
@@ -1,78 +1,34 @@
...
+    /* Enable strict null checks. */
+    "esModuleInterop": true,
+    /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */
+    "experimentalDecorators": true,
+    /* Enables experimental support for ES7 decorators. */
+    "emitDecoratorMetadata": true,
+    /* Enables experimental support for emitting type metadata for decorators. */
+    "outDir": "build",
+    "forceConsistentCasingInFileNames": true,
+    /* Disallow inconsistently-cased references to the same file. */
```
geisterfurz007, line 21: The comments here moved around due to formatting and describe the things above them.

package.json
```diff
@@ -35,6 +40,7 @@
     "winston": "3.3.3"
   },
   "devDependencies": {
+    "@tsconfig/node14": "1.0.1",
```
geisterfurz007, line 43: This isn't used (if I understand correctly)

geisterfurz007 (changes requested):

docker-compose.yml
```diff
@@ -9,3 +9,19 @@ services:
...
+    depends_on:
+      - postgres
```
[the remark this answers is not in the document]
adrian-goe: Why should I have to specify a version? Postgres is using version 13 and yes bot is using the latest one that will be deployed in GitHub Container Registry.

I can create a seperat file.
But if you only want to have the database, you can run `docker-compose up -d postgres` this starts just the posgres container

tsconfig.json
adrian-goe: I would remove them all. The docs are good enogh. 🙈

docker-compose.yml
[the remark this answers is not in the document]
geisterfurz007: In development you will almost always want just the database so I thought it made sense to have that as default. Then again it's true, it's just one more word. This would require an update in the [CONTRIBUTING.md](https://github.com/Yes-Theory-Fam/yesbot-ts/blob/master/CONTRIBUTING.md#run-the-database) then.

tsconfig.json
geisterfurz007: 👍

docker-compose.yml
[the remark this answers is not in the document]
adrian-goe: I updated the CONTRIBUTING.md 👍🏻

geisterfurz007 (approved):

geisterfurz007:
LGTM!

PR #285, opened 2021-06-20
fix bug with false mentions

adrian-goe:
There was a false. Instead of `@username` it was `<@@username>`
getting the ID should fix this.


geisterfurz007:
The same issue is present as well here:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/4607c5c5a4623f5c322bdb79cc4c11255d91a1a1/src/programs/voice-on-demand.ts#L355-L357

geisterfurz007 (approved):

PR #287, opened 2021-06-21
add test for extractMessageInfo

adrian-goe:
Add test for `extractMessageInfo()` in message.ts from event_distribution


geisterfurz007 (approved):
LGTM!

PR #289, opened 2021-06-21
Feature/deadchat for groups

hasanlatch007:
Also includes deleting `.npmrc` file, uses the same function as !group changeCooldown for moderations.
Pretty straight forward, reminder to merge the db update so you don't get a bunch of red lines :)!


hasanlatch007:
closes #142

src/programs/deadchat-groups.ts
```diff
@@ -0,0 +1,61 @@
+import { Message } from "discord.js";
```
geisterfurz007, line 1: Technically this file doesn't contain a single program. It should either be moved to group-manager.ts or a new group-manager folder should be created to hold everything relevant to groups.

src/programs/deadchat-groups.ts
```diff
@@ -0,0 +1,61 @@
...
+import Tools from "../common/tools";
+import prisma from "../prisma";
+
+export const deadchatGroup = async (
```
geisterfurz007, line 5: The name is not ideal. It returns whether a chat is dead enough for a group so something like `isChatDeadForGroup` or `isChatDead` would be a better fit (this includes the filename as well).

src/programs/deadchat-groups.ts
```diff
@@ -0,0 +1,61 @@
...
+      message,
+      `Chat is not dead! You can ping this group in the next ${timeRemaning} minutes.`
+    );
+    return false;
+  }
+
+  if (isDead) {
+    return true;
+  }
+};
```
geisterfurz007, line 59: The side-effects in this function are not ideal. The only thing changing in this large block is the output based on the time remaining till dead. Maybe instead of a boolean + side-effects, this function could return the remaining minutes or seconds till the chat would be dead and the side-effects happen at the callsite of this? In this case the name should be changed as well (ignoring my earlier comment).

The text could be boiled down to \`Chat is not dead! You can ping this group in the next ${timeRemainingString}\` to avoid repetitions (and sneaky bugs where the error happens twice which happens to be the case here if `timeRemaining === 1` as there is no return after it ;) ).

src/programs/group-manager.ts
```diff
@@ -716,7 +764,7 @@ const isChannelAllowed = (channel: Channel): boolean => {
   const allowedCategories = ["hobbies", "gaming"];
   const allowedChannels = [
     "449984633908625409", // chat
-    "623565093166252052", // chat-too
+    "856479423490031640", // chat-too
```
geisterfurz007, line 767: The ID should not be changed.

geisterfurz007 (changes requested):

geisterfurz007 (changes requested):
Update `@yes-theory-fam/database` to `0.6.1` which contains the deadtime field on UserGroup.

src/programs/group-manager.ts
```diff
@@ -174,8 +173,13 @@ const groupManager = async (message: Message, isConfig: boolean) => {
 
     const group = matchingGroups[0];
     const timeDifference = (Date.now() - group.lastUsed.getTime()) / 1000 / 60;
+    const timeRemaningforDeadchat = await isChatDead(message, group.name)
```
findtravishere, line 176: `timeRemainingForDeadchat`

hasanlatch007:
Updated my typo :'). Did not go for seconds the idea of it saying `1 minutes` does bother me but it is what it is, code works as functioned hopefully my variable names are creative enough for this one :')

src/programs/group-manager.ts
```diff
@@ -174,8 +173,13 @@ const groupManager = async (message: Message, isConfig: boolean) => {
 
     const group = matchingGroups[0];
     const timeDifference = (Date.now() - group.lastUsed.getTime()) / 1000 / 60;
+    const timeRemainingForDeadchat = await isChatDead(message, group.name)
 
-    if (!(await deadchatGroup(message, group.name))) {
+    if (timeRemainingForDeadchat >= group.deadtime) {
```
geisterfurz007, line 178: I might be incorrect because it's late but I believe this is wrong. `timeRemainingForDeadchat` contains the time chat needed to be dead for to allow a ping. So any value **greater than 0** means the ping is not allowed, not any value greater than the group's dead time.

Given that `timeRemainingForDeadchat` is (essentially) guaranteed to be less than `group.deadtime` (since the function subtracts from it), this is never true, is it? I will try this out tomorrow when I am not 30 minutes late to sleep and report back.

src/programs/group-manager.ts
```diff
@@ -783,4 +787,29 @@ const isChannelAllowed = (channel: Channel): boolean => {
   return allowedChannels.includes(channel.id);
 };
 
+const isChatDead = async (
```
geisterfurz007, line 790: The function name isn't accurate anymore since it doesn't return a boolean but a number.

src/programs/group-manager.ts
```diff
@@ -783,4 +787,29 @@ const isChannelAllowed = (channel: Channel): boolean => {
...
+  requestedGroup: string
+) => {
+  const group = await prisma.userGroup.findFirst({
+    where: {
+      name: {
+        equals: requestedGroup,
+        mode: "insensitive",
+      },
+    },
+  });
```
geisterfurz007, line 801: You can save the additional database query by passing in the group instead of the group name.

geisterfurz007 (changes requested):

src/programs/group-manager.ts
hasanlatch007: damn I just saw this after committing :')

src/programs/group-manager.ts
hasanlatch007: For example, let's take this if `deadtime` is equal to 5, and `Date. now()` is bigger than the last messages send `timeDifference` will be a positive number if it's bigger than the deadtime, `timeRemaining` will return a negative number.
Resulting in the chat being able to be pinged, testing went well for this, but I could have my math wrong for this, so I won't commit anymore before you've checked it.

geisterfurz007: I double checked this and it seems broken indeed; to reproduce:

1. Create a new group
2. Set the deadtime to 15 minutes
3. Send a message in #music in the hobbies section
4. Wait 3 minutes
5. Ping the group

Expected: The chat is not dead (since you pinged it 3 minutes ago) so there is no ping
Actual: The group is pinged.

`if (timeRemainingForDeadchat > 0) {` appears to work as intended though!

geisterfurz007:
@yes-theory-fam/database still needs to be updated to 0.6.1 if I am not mistaken

src/programs/group-manager.ts
```diff
@@ -496,6 +520,37 @@ const updateGroup = async (
   );
 };
 
+const changeDeadtime = async (
+  message: Message,
+  requestedGroupName: string,
+  newDeadtime: string
+) => {
+  const deadtimeNumber = Number(newDeadtime);
+  if (isNaN(deadtimeNumber)) {
```
sklirg, line 529: What happens if the number is negative? What should happen?

src/programs/group-manager.ts
```diff
@@ -496,6 +520,37 @@ const updateGroup = async (
...
+  }
+
+  const group = await prisma.userGroup.findFirst({
+    where: {
+      name: {
+        equals: requestedGroupName,
+        mode: "insensitive",
+      },
+    },
+  });
```
sklirg, line 544: This should handle the case of the group not being found.

src/programs/group-manager.ts
```diff
@@ -496,6 +520,37 @@ const updateGroup = async (
...
+
+  await prisma.userGroup.update({
```
sklirg, line 546: This should handle the case of the update failing.

src/programs/group-manager.ts
```diff
@@ -746,6 +801,21 @@ const isChannelAllowed = (channel: Channel): boolean => {
...
+    await message.channel.messages.fetch({ limit: 2 })
+  ).array();
+  let timeDifference = deadTime;
+
+  if (lastMessages.length === 2) {
+    timeDifference =
+      (Date.now() - lastMessages[1].createdTimestamp) / 1000 / 60;
+  }
+
+  return deadTime - Math.round(timeDifference);
```
geisterfurz007, line 816: This might be picky but I found this hard-ish to read. Instead of defaulting to deadTime for the case lastMessages.length is *not* 2 to then return deadTime - deadTime later, it would be more straightforward to just return 0 in that case:

```ts
if (lastMessages.length < 2) {
  return 0;
}

const timeDifference = (Date.now() - lastMessages[1].createdTimestamp) / 1000 / 60;
return deadTime - Math.round(timeDifference);
```

sklirg, line 816: The assignment of `deadTime` and `timeDifference` here can be confusing.

I'd suggest _not_ assigning `const deadTime = group.deadtime`, and rather use `return group.deadtime - ...`.

src/programs/group-manager.ts
```diff
@@ -746,6 +801,21 @@ const isChannelAllowed = (channel: Channel): boolean => {
...
+const timeRemainingForDeadchat = async (message: Message, group: UserGroup) => {
+  const deadTime = group.deadtime;
+  const lastMessages = (
+    await message.channel.messages.fetch({ limit: 2 })
+  ).array();
+  let timeDifference = deadTime;
+
+  if (lastMessages.length === 2) {
+    timeDifference =
+      (Date.now() - lastMessages[1].createdTimestamp) / 1000 / 60;
```
sklirg, line 813: This could return early instead of assigning and then returning after the conditional.

src/programs/group-manager.ts
sklirg, line 816: What should happen if the conditional fails? Currently, it returns the `group.deadtime` value that might always be above 0 (if the group has a deadtime configured).

src/programs/group-manager.ts
```diff
@@ -543,10 +543,20 @@ const changeDeadtime = async (
...
+
+  try {
+    await prisma.userGroup.update({
+      where: { id: group.id },
+      data: { deadtime: deadtimeNumber },
+    });
+  } catch (error) {
+    logger.error("Failed to update database group deadTime," + error);
+    await message.react("👎");
+  }
```
geisterfurz007, line 559: There should be a return after this (otherwise the bot will add 👍 as well)

geisterfurz007 (changes requested):

src/programs/group-manager.ts
hasanlatch007: Bot is bipolar let it be :')

geisterfurz007 (approved):

PR #299, opened 2021-06-24
transition for new country roles

adrian-goe:
New way of checking the country role

### DIFF
--- a/src/utils/country-role-finder.ts
+++ b/src/utils/country-role-finder.ts
@@ -1,27 +1,46 @@
-export class CountryRoleFinder {
-  private static rollRegex = /([\S ]*?) (:flag[a-z_]*?:)/g;
+import { countries, Country } from "../collections/flagEmojis";
+import { Role } from "discord.js";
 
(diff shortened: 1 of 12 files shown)


src/utils/country-role-finder.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  private static getMatches(input: string): Country {
+    return countries.find((country) => {
+      return (
+        input === `I'm from ${country.name}` ||
+        input === `I'm from ${country.name}!` ||
+        input === `I'm from ${country.name}! ${country.emoji}` ||
+        input === `${country.name} ${country.emoji}`
+      );
+    });
```
geisterfurz007, line 44: These checks all look very similar; we should move that logic to a function.

geisterfurz007 (changes requested):

src/utils/country-role-finder.ts
geisterfurz007: Addressed in #305

geisterfurz007 (approved):

PR #300, opened 2021-06-24
DRAFT: Remove old countryrole checks

adrian-goe:
⚠️ this should not merged right away ⚠️

Remove old country-role checks after transition is over.

This is the follow-up PR for https://github.com/Yes-Theory-Fam/yesbot-ts/pull/299

### DIFF
--- a/src/utils/country-role-finder.ts
+++ b/src/utils/country-role-finder.ts
@@ -1,0 +1,27 @@
+export class CountryRoleFinder {
+  private static rollRegex = /([\S ]*?) (:flag[a-z_]*?:)/g;
+
+  static getCountryByRole(input: string): string {
+    const result = this.getMatches(input);
+    return result[1];
+  }
+
+  static isCountryRole(input: string): boolean {
+    const result = this.getMatches(input);
+    return !!result.length;
+  }
+
+  private static getMatches(input: string) {
+    let match;
+    let result: string[] = [];
+    while ((match = CountryRoleFinder.rollRegex.exec(input)) !== null) {
+      if (match.index === CountryRoleFinder.rollRegex.lastIndex) {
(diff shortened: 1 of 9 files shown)


src/utils/country-role-finder.ts
```diff
@@ -0,0 +1,27 @@
+export class CountryRoleFinder {
+  private static rollRegex = /([\S ]*?) (:flag[a-z_]*?:)/g;
```
sklirg, line 2: The flags are actual UTF-8 emojis, not the string values of `:flag_XX:`. This makes the test case incorrect as well, as it should expect an emoji and not `:flag_de:`.

Example: `Germany 🇩🇪`

sklirg, line 2: If you don't need the first group for anything (`return result[1]`), you shouldn't need to capture it as a group, or you can mark it as a non-capturing group with `(?:regex)`.

src/utils/country-role-finder.ts
```diff
@@ -0,0 +1,27 @@
...
+    const result = this.getMatches(input);
+    return !!result.length;
+  }
+
+  private static getMatches(input: string) {
+    let match;
+    let result: string[] = [];
+    while ((match = CountryRoleFinder.rollRegex.exec(input)) !== null) {
+      if (match.index === CountryRoleFinder.rollRegex.lastIndex) {
+        CountryRoleFinder.rollRegex.lastIndex++;
```
sklirg, line 19: What is this used for?

src/utils/country-role-finder.ts
```diff
@@ -0,0 +1,27 @@
...
+      }
+      match.forEach((match) => {
```
sklirg, line 21: Why are we looping over every match and pushing it to a new list? Can we just do `return match` or `return [...match]`?

PR #305, opened 2021-06-24
Simplify the country role finder

geisterfurz007:
Addresses https://github.com/Yes-Theory-Fam/yesbot-ts/pull/299#discussion_r658131187 and simplifies the check by a bit to cover more possible cases.

### DIFF
--- a/src/utils/country-role-finder.ts
+++ b/src/utils/country-role-finder.ts
@@ -1,6 +1,8 @@
 import { countries, Country } from "../collections/flagEmojis";
 import { Role } from "discord.js";
 
+type FinderCountryProperties = Pick<Country, "name" | "emoji">;
+
 export class CountryRoleFinder {
   static getCountryByRole(input: string): string {
     const result = this.getMatches(input);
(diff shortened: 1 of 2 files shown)


adrian-goe (approved):

src/utils/country-role-finder.ts
```diff
@@ -18,29 +20,19 @@ export class CountryRoleFinder {
...
-        input === `I'm from ${country.name}! ${country.emoji}` ||
-        input === `${country.name} ${country.emoji}`
-      );
-    });
+    return countries.find((country) => this.check(country, input));
+  }
+
+  private static check(country: FinderCountryProperties, compare: string) {
+    return (
+      compare.includes(country.emoji) || compare === `I'm from ${country.name}!`
```
adrian-goe, line 35: Well, this just work for current roles right.
would add the future roll check as well

geisterfurz007: Future roles should be covered by `compare.includes(country.emoji)`, shouldn't they? Technically that's one for the mod manual that all new roles are required to contain their countries flag in the name.

Issue #302, opened 2021-06-24
Replace channel IDs with channel names for group pings


geisterfurz007:
Groups are currently permitted by channel ID which is prone to accidental commits for trying group related features out on a test-channel:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/68dcbbf3998d256de7913d2db8263d47c8dfefac/src/programs/group-manager.ts#L717-L726

This can be avoided by replacing them with their channel names and comparing against that.

geisterfurz007:
Closed in #307

PR #307, opened 2021-06-24
Chore: replace hard coded IDs for Group Pings

hasanlatch007:
for chore #302 
Hopefully, I got the idea correct


adrian-goe:
The idea is correct. But I would prefer an enum with the channel names, so all needed channels are in the same file.

src/programs/group-manager.ts
```diff
@@ -715,15 +715,33 @@ const isChannelAllowed = (channel: Channel): boolean => {
...
+    ).id,
+    channel.guild.channels.cache.find(
+      (channel) => channel.name === ChatNames.CHAT_FIVE
+    ).id,
+    channel.guild.channels.cache.find(
+      (channel) => channel.name === ChatNames.SELF_DEVELOPMENT
+    ).id,
+    channel.guild.channels.cache.find(
+      (channel) => channel.name === ChatNames.DAILY_CHALLENGE
+    ).id,
```
geisterfurz007, line 744: Instead of getting all channel IDs here, you can list the chat names here and change the return value of the function further down to `allowedChannels.includes(channel.name)`.

geisterfurz007 (changes requested):

geisterfurz007 (approved):

geisterfurz007:
Technically we can replace all the channel names in message.ts as well but since we want to tear the whole thing down anyway, I think it makes sense to use the enum when we move to the new event distribution 👍 
LGTM!

PR #311, opened 2021-06-25
fix/country role related commands

geisterfurz007:
- fix: double pings when selecting region role
- fix: !profile for US members
- fix: timezones by country for !birthday

The double pings were caused by a faulty condition that would trigger in the event where the member gets their general role removed since in that case oldMember has it and newMember has a region role. This was fixed by checking if newMember has more roles than oldMember. This fix has been live through the night and appears to work properly.

!profile was fixed by adding an optional boolean to the CountryRoleFinder that allows regional roles to be picked up which is a requirement for both !profile and !birthday.

In addition to that !birthday contained some manual overrides for `the USA` (amongst others) which now failed as the `the` is not part of the country string anymore.


adrian-goe (approved):
LGTM

PR #316, opened 2021-06-26
Feature: ping yestheoryuploads group

hasanlatch007:
This is very basic and does what the basic idea wants it to do until we implement the new events, this should only be merged if we wanted to role out the feature early or when Adrian finish reworking it to use the new events


src/events/message.ts
```diff
@@ -203,6 +205,8 @@ const routeMessage = async (message: Message) => {
     await GroupManager(message, true);
 
   if (words.includes("@group")) await GroupManager(message, false);
+  if (channel.name === ChatNames.YESTHEORY_POSTED)
+    await YesTheoryUploadedPing(message); //as long as no support sends anything in that channel this won't be called
```
hasanlatch007, line 209: This comment has been fixed and can be ignored

adrian-goe: Then remove the comment when its fixed 🙈

adrian-goe (changes requested):

adrian-goe (approved):

Issue #320, opened 2021-06-26
Support should be able to decide if a group is user pingable


adrian-goe:
Some groups like `dailychallenge` should not be allowed to be pingable by users other than, e.g. support.
Therefore, the group command should be extended with this new feature. A database update is needed.

Supports must also have the option to change this decision later.

It is an option to not only allow specific roles to ping a group, but also just the group itself.

PR #323, opened 2021-06-27
fix: disable group YesTheoryUploads for members

geisterfurz007:
We do not want the YesTheoryUploads group to be pinged by all members, just by the bot (ignoring even supports and admins for now).

Until #320 is closed, a little string comparison shall do the trick.


hasanlatch007 (approved):
LGTM!

PR #324, opened 2021-06-27
Feature: timeout command

hasanlatch007:
As requested a time out command for support and companion only, all logging goes to #bot-output


geisterfurz007 (approved):
LGTM, thanks a bunch!

PR #326, opened 2021-06-28
Issue: voice room claim

hasanlatch007:
This should fix the allowedList always returning empty, also added in case the rooms need to be deleted to also delete the old host from prisma.


geisterfurz007 (approved):
LGTM!

Issue #325, opened 2021-06-27
Bot does not take account leap years for birthdays


hasanlatch007:
`!birthdays feb-29` will return march 1, this is due to typescript library not taking leap years into account in the `getmonth()`, this should be fixed if we take this function into account as it will return false or true if the current year is a leap year

> function isDateRangeEqualToLeapYear(year)
{
  return !((year % 4) && (year % 100) || !(year % 400));
}

PR #329, opened 2021-06-28
#325 fix: accept feb-29th as birthday

geisterfurz007:
The bot previously used the year 1970 to return the date so feb 29th automatically rolled over to march 1st. Instead we can use year 1972 which was a leapyear to return and save the date.


adrian-goe:
Why not using the year the people give?

geisterfurz007:
Because people don't give us a year and we don't really care about the age of people, so we don't need the year.

adrian-goe (approved):

PR #346, opened 2021-07-02
feat: ready event

geisterfurz007:
This PR adds support for the ready event to the new event distribution system. It also adds a few tests that were forgotten in an earlier PR.

- feat: Add GuildMemberUpdate to event distribution
- feat: Ready event in event distribution

### DIFF
--- a/__tests__/event-distribution/events/events.spec.ts
+++ b/__tests__/event-distribution/events/events.spec.ts
@@ -5,19 +5,27 @@ import {
 import {
   addMessageHandler,
   extractMessageInfo,
-  MessageEventHandlerOptions,
 } from "../../../src/event-distribution/events/message";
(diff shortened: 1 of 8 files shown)


adrian-goe:
closes #261

adrian-goe:
`console.log("Called handler 6");` is not called on start.

adrian-goe (approved):
I would suggest also change in the index file line 26 to:
```ts
const bot = new Client({ partials: ["REACTION", "MESSAGE"] });
logger.info("Initializing event-distribution");
distribution.initialize().then(async () => {
  logger.debug("Logging in to Discord Gateway");
  await bot.login(process.env.BOT_TOKEN);
});
```
otherwise, it looks good 😊

Issue #255, opened 2021-06-16
Deploy the bot as Docker Application


adrian-goe:
Auto deploy the bot with GitHub action.
* on master as preview instance
* on tag as production instance

adrian-goe:
this issue is blocked by #254

sklirg:
Not sure if this fits exactly into this issue, but anyways:

Update github actions with different steps and allow them to run on PRs/pushes.

Examples:
- linting, should not require installing a private npm package - can be run on PR
- (unit) testing, should not require installing a private npm package - can be run on PR
- (integration) testing (prisma), depends on installing a private npm package - should not be run on PR (since this can expose the github secret used in installing the package)

adrian-goe:
Partially done with:

- https://github.com/Yes-Theory-Fam/yesbot-ts/pull/330
- https://github.com/Yes-Theory-Fam/yesbot-ts/pull/343
- https://github.com/Yes-Theory-Fam/yesbot-ts/pull/344
- https://github.com/Yes-Theory-Fam/yesbot-ts/pull/345

adrian-goe:
Need to add the deployment script

PR #347, opened 2021-07-02
Ci add deploy action

adrian-goe:


adrian-goe:
Relates to #255

deployment/updater.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+if [[ -f databasebackup.sh ]];
+ then
+	echo backing up bot
+	./databasebackup.sh
```
geisterfurz007, line 8: Is this file missing?

adrian-goe: I can add a sample file, but this file contains the database password and is on the server in the `yesbot` dir

geisterfurz007: That's a very valid point! I could use pg_dump without password as yesbot. Would it make sense to configure the postgres container so that we can do the same despite docker (also @sklirg)? Otherwise I think an example version of the script for reference would be a good thing to have either way.

adrian-goe: I guess you can do this for your database, but not for the one in the container. I added the sample.

sklirg: Postgres can allow user accounts to log in to the database if configured, but that trusts the linux user account system. Since the database runs in docker (i.e. not on the host itself), directly allowing it to log in, doesn't work.

geisterfurz007 (approved):

PR #353, opened 2021-07-03
feat(*): send automatic release notes on start

adrian-goe:
the bot can now find the commit-messages since the last deployment and
send them into the update channel

this is blocked by https://github.com/Yes-Theory-Fam/database/pull/18


adrian-goe:
This is how it would look like. The first one is a start with a new tag and the second one is for a reboot without any update
Screenshot 2021-07-04 at 00 49 02

src/programs/release-notes.controller.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  handle(client: Client): void {
+    this.githubReleaseNotesService.handle().then((message) => {
+      if (!!message.tagMessage) {
+        const updateChannel = client.guilds
+          .resolve(process.env.GUILD_ID)
+          .channels.cache.find((channel) => channel.name === ChatNames.UPDATES);
+        // @ts-ignore
+        updateChannel.send(message.tagMessage);
```
geisterfurz007, line 25: You can avoid the `@ts-ignore` by using `as TextChannel` after the find.

src/programs/release-notes.controller.ts
```diff
@@ -0,0 +1,45 @@
...
+      }
+
+      if (!!message.releaseNotes) {
+        const botDevChannel = client.guilds
+          .resolve(process.env.GUILD_ID)
+          .channels.cache.find(
+            (channel) => channel.name === ChatNames.BOT_DEVELOPMENT
+          );
+        // @ts-ignore
+        botDevChannel.send(message.releaseNotes);
```
geisterfurz007, line 37: You can avoid the `@ts-ignore` by using `as TextChannel` after the find.

src/service/github-release-notes.service.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  private createReleaseNotes(releaseNotes: string) {
+    return `Hello everyone :wave: I'm back after a short break :sleeping:
+During this time I have been worked on a lot. Here is a small overview:
+
+${releaseNotes}
+So now back to work :tools:     
+`;
+  }
+}
```
geisterfurz007, line 106: WebStorm reports this as unused.

src/service/github-release-notes.service.ts
```diff
@@ -0,0 +1,106 @@
...
+        new Date(a.node.publishedAt).getTime() -
+        new Date(b.node.publishedAt).getTime()
+      );
+    });
+    const oldRelease = sortedData[0];
+    const newRelease = sortedData[1];
+    return { oldRelease, newRelease };
+  }
+
+  private async getCommitMessages(lastRelease: any, newRelease: Edge) {
```
geisterfurz007, line 65:
```suggestion
  private async getCommitMessages(lastRelease: Release, newRelease: Edge) {
```

yarn.lock
```diff
@@ -888,10 +888,10 @@
...
     "@types/yargs-parser" "*"
 
-"@yes-theory-fam/database@0.7.1":
-  version "0.7.1"
-  resolved "https://npm.pkg.github.com/download/@yes-theory-fam/database/0.7.1/1700488c110118ae3695399d8780dc4ca47dafc7fd7f2aaf354fd2250c0cbeb9#08ad1299e31bd0fb4454b22073c6f071b5e7d000"
-  integrity sha512-k2CspXXsIZcN7yJ2BQzTW31X4PlHFL9jvz6d06ZS4uvE9wuIct5AnHnzJePt9ow25pBJud1adi+LFruKXBWlQg==
+"@yes-theory-fam/database@0.8.0":
+  version "0.8.0"
+  resolved "https://npm.pkg.github.com/download/@yes-theory-fam/database/0.8.0/b3ce43e34ebc54091c5e49f16692209d9b8b380fca7cc46d3656e5425fa4797f#d1cce41b66a527b5ff99e5c784e95278175c3425"
+  integrity sha512-UQWe9RwqrHBUoprs3RbmKaXAiqdrfDzwwk/X3chcJfpwSD5ATv3epn2UZa7YbJPjB0AmLtoGM2eZe1sdET1Tyg==
```
geisterfurz007, line 894: This will have to be regenerated, now that 0.9.0 is out.

geisterfurz007 (changes requested):

geisterfurz007:
I am a strong advocate for feature folders instead of having folders for each type of file (like service and accessor which were introduced in this PR). In development this means that you have to have three different directories open in the project explorer with potentially a dozen other files in there which have absolutely nothing to do with what you are working on, atm.

Instead, I think it would make sense to have `src/programs/release-notes/` (or `github-release-notes`) which would then contain all bits required for that feature to work (i.e. the types, API call to GitHub, service to handle the database and controller if we shall stick to the name for communication to Discord). That way if there was any work required on the feature, I have one folder I need to open and have everything needed to work on it.

This setup would also mean that a bunch of extra .ts / .js files are loaded by the event distribution. We could mitigate that by using a more specific file extension like .controller.ts which you already started using here to make sure only relevant files are picked up by the glob pattern (although I am not sure loading a couple more files on startup will have that much impact on startup time at the small scale of YesBot).

adrian-goe:
I kinda agree with you but on the other side not. The reason is a Clean Architecture approach where you have layers.
If all the files related to one feature set in one directory would mean a strong coupling with potential a lot of rewritten code.
By using a folder for a layer, we can reuse the GitHub accessor for other things.
In the layers, I agree we can sort this in directories, so it isn't that overwhelming.

__tests__/programs/release-notes/usecase/github-release-notes.usecase.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    githubReleaseNotesUsecase = new GithubReleaseNotesUsecase();
+
+    githubReleaseNotesAccessor = new GithubReleaseNotesAccessor();
+    githubReleaseNotesAccessor.getCommitMessages = jest.fn(() =>
+      Promise.resolve(getCommitMessagesResponse)
+    );
+    githubReleaseNotesUsecase["githubReleaseNotesAccessor"] =
+      githubReleaseNotesAccessor;
+  });
+  it("should sort CommitMessages", () => {
```
geisterfurz007, line 43: Should sort releases?

geisterfurz007 (commented):
Rest LGTM for now; pipeline still disagrees though

geisterfurz007 (approved):
LGTM!

PR #361, opened 2021-07-11
fixed yestheoryuploaded ping

hasanlatch007:
If webhooks count as bots this should not return an error and we should no longer need to use `hasRole`.
Used the test YesBot by forcing it to send a message in #yestheory-uploaded which pinged the group.

### DIFF
--- a/src/programs/yestheory-content.ts
+++ b/src/programs/yestheory-content.ts
@@ -1,25 +1,21 @@
 import Discord, { Message, TextChannel } from "discord.js";
 import { ChatNames } from "../collections/chat-names";
-import { hasRole } from "../common/moderator";
 
 //keep in mind this is very temporarily basic code, until we have the new events up and ready
 const YesTheoryUploadedPing = async (message: Message) => {
-  const user = message.author.id;
-  const MemberGuild = await message.guild.members.fetch(user);
-  if (hasRole(MemberGuild, "Support")) {
-    return;
+  if (message.author.bot) {
+    const channelDiscussion = message.guild.channels.cache.find(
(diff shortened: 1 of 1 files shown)


hasanlatch007:
This actually might not work I checked the discord.js docs and webhooks are not logged in as bots? So `message.author.bot` might always return false, I'm not sure of this I gotta read the docs a bit better.
`if (message.webhookID)` Will have to replace it with this I think :')

hasanlatch007:
Both `webhookID` or `message.author.bot` work but the first one will only return true if the message is from a webhook that serves our goal better.

src/programs/yestheory-content.ts
```diff
@@ -1,25 +1,21 @@
...
 import { ChatNames } from "../collections/chat-names";
-import { hasRole } from "../common/moderator";
 
 //keep in mind this is very temporarily basic code, until we have the new events up and ready
 const YesTheoryUploadedPing = async (message: Message) => {
-  const user = message.author.id;
-  const MemberGuild = await message.guild.members.fetch(user);
-  if (hasRole(MemberGuild, "Support")) {
-    return;
+  if (message.webhookID) {
```
geisterfurz007, line 6: Use an early return instead.

geisterfurz007 (changes requested):

geisterfurz007:
`webhookID` was a really nice catch!

src/programs/yestheory-content.ts
hasanlatch007: Done :)

hasanlatch007:
> `webhookID` was a really nice catch!

I used my big brain for once :D

geisterfurz007 (approved):

PR #362, opened 2021-07-15
Feat: Option for group ping

hasanlatch007:
Welp, I got bored, and decided to work on this, probably needs lots of feedback, legit coded this in 30 minutes, works as intended but I feel like I could improve it better.


src/programs/group-manager.ts
```diff
@@ -66,6 +66,8 @@ const groupManager = async (message: Message, isConfig: boolean) => {
         "toggle",
         "changeCooldown",
         "changeDeadtime",
+        "changeGroupPingSettings",
+        ``,
```
geisterfurz007, line 70: Empty string should be removed.

src/programs/group-manager.ts
```diff
@@ -562,6 +601,54 @@ const changeDeadtime = async (
...
+
+  if (!group) {
+    await message.reply("That group doesn't exist!");
+    return;
+  }
+
+  try {
+    await prisma.userGroup.update({
+      where: { id: group.id },
+      data: { groupPingOption: setting },
```
geisterfurz007, line 641: `groupPingOption` does not exist in the current database schema.

Available options should also be [defined as enum](https://www.prisma.io/docs/concepts/components/prisma-schema/data-model#defining-enums). I believe that will also generate them as TS so the string comparisons further above should be replaced with those enums.

geisterfurz007 (changes requested):

src/programs/group-manager.ts
hasanlatch007: This was resolved my bad!

src/programs/group-manager.ts
hasanlatch007: I created a PR on the database github, all options have been defined as enum

hasanlatch007:
If this is good to go dont merge it till i delete the function michel added for the group YesTheoryUploaded

src/programs/group-manager.ts
```diff
@@ -562,6 +601,54 @@ const changeDeadtime = async (
...
+  option: string
+) => {
+  const setting = option.toUpperCase();
+
+  if (
+    (setting !== "MODERATOR" &&
+      setting !== "MEMBER" &&
+      setting !== "BOT" &&
+      setting !== "OFF") ||
+    !setting
```
geisterfurz007, line 616: - This should use the enum values as well
-  `|| !setting` is redundant because an empty string or undefined won't match any of the previous cases either.

geisterfurz007 (changes requested):
- Update @yes-theory-fam/database to 0.11.0
- Remove this: https://github.com/Yes-Theory-Fam/yesbot-ts/blob/dafd7cb319bb834d37bacc658aa32b9c4456b928/src/programs/group-manager.ts#L916-L919

geisterfurz007 (approved):

Issue #369, opened 2021-07-20
Time Out does not deny channel view access


geisterfurz007:
Permissions granted through roles always outweigh permissions denied through roles (no matter the index in the role list).

The code that handles this, is around here

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/5da081f6a4cad15ab16bbf2ae42b151e5bc54e7a/src/events/guild-member-update.ts#L75-L90

and doesn't function anymore.

hasanlatch007:
Looked around a bit, the server permissions for each channel for `Time Out` allows viewing the channel, so technically this code does nothing, turn off the bot assign someone the Break role and it will work normally. I'm guessing this function is made to delete user permissions to channels like hobby and others? If not then I have no clue what this function does :')

geisterfurz007:
Oh, you are absolutely right! I did link to the wrong snippet in the issue, it does exactly what you said it does. The code that locks members out of their country channels is only run on Break role, not Time Out:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/5da081f6a4cad15ab16bbf2ae42b151e5bc54e7a/src/events/guild-member-update.ts#L30-L32

In that case everything might work as intended with the current bot implementation and it was just a permission issue on the server. I will double check that later when I am back home.

hasanlatch007:
Added the `lockCountryChannels` and `unlockCountryChannels` for time out roles you can find it here #373 :)

geisterfurz007:
I double checked the permissions on the server and everything is working as intended; sorry for prematurely opening this as issue.

PR #373, opened 2021-07-21
Fix: Lock users with timeout role from country channels

hasanlatch007:
This won't fix issue #369 since it seems to be server permission issues but should fix it to lock users with the `Time Out` role out of country channels


geisterfurz007:
This turned out to not be needed; see https://github.com/Yes-Theory-Fam/yesbot-ts/issues/369#issuecomment-885938810

PR #387, opened 2021-08-01
feat: timer events and quote voting

geisterfurz007:
This PR introduces two larger changes:

### Timer events

They are still listed in the DiscordEvents eventhough that's not fully accurate, to leverage the existing event hub structure.  
Timer events are split into two pieces:
- A service with static methods to schedule and run timers which are persisted in the database
- An event handler to load and schedule all timers on bot ready event

### Quote voting
To help the quality of quotes in the quotes channel, YesBot will add two reactions on new posts which are tallied after 24 hours. If the ratio of positive to negative is below 4, the post will be removed. If it is above 9, the message will be pinned and the system message about the pinned post will be removed.


geisterfurz007:
Blocked by https://github.com/Yes-Theory-Fam/database/pull/31

adrian-goe (approved):

PR #389, opened 2021-08-01
Support for GitHub mentions -> Discord mentions in release notes

sklirg:
I found it hard to read the release notes on GitHub when they contained mentions of people using numerical IDs, such as `312297787901607937`. This feature allows for conversion from GitHub username mentions (such as @sklirg) to Discord mentions by supplying a static file with the conversion (`src/programs/release-notes/config/github-to-discord-names.ts`).

It will keep discord mentions for people not on GitHub as they are (text such as `<@1234>` will stay as `<@1234>`), so it is possible to Discord-mention someone without them having a GitHub account. And people not present in the conversion file will get "mentioned" by their GitHub username instead.

I also added a default for the GitHub GraphQL API URL.


geisterfurz007:
I might be mistaken but I think the function `createReleaseNotes` creates the string sent in bot-development:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/fe88364412304b68ef4a13a2aad7b7b2b8652bbf/src/programs/release-notes/usecase/github-release-notes.usecase.ts#L43-L44

`releaseNotes` would be the one with the modified pings however `tagMessage` is the one sent in updates:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/fe88364412304b68ef4a13a2aad7b7b2b8652bbf/src/programs/release-notes/controller/release-notes.controller.ts#L26-L31

adrian-goe:
that's correct @geisterfurz007

PR #395, opened 2021-08-06
Fix: detect multiple flags

### LINKED ISSUE #382: Bot incorrectly detects multiple flags
Sometimes when members drop their flag multiple times without spaces, the bot incorrectly registers multiple flags:

grafik

Cause of this is the detection of flags which is done by a simple string lookup. In the case in the image above, the indicators are `INININ` which also makes the flag of Nicaragua (NI) match.

hasanlatch007:
This should close #382, the way emojis are on discord is combining multiple emojis together (does this even make sense..?). The current regex splits them up separately, combines the first 2 indexes and the last 2 in the array, which together will give the first 2 flags send. If they are different, the bot replies with the message. This kind of can be bypassed by sending the same flag twice then a different one but either way he will still get assigned the role of the first flag detected, but the bug was mostly having the same flag twice and it gets detected as different, so that doesn't really matter.

### DIFF
--- a/src/programs/where-are-you-from.ts
+++ b/src/programs/where-are-you-from.ts
@@ -21,7 +21,9 @@ const whereAreYouFrom = async (message: Message) => {
     );
 
     if (matchedCountries.length === 0) {
-      await message.reply("You must send the flag of the country you are in!");
+      await message.reply(
+        "You must send the flag or name of the country you are in!"
+      );
       return;
     }
 
--- a/src/programs/where-are-you-from.ts
+++ b/src/programs/where-are-you-from.ts
@@ -20,7 +20,7 @@ const whereAreYouFrom = async (message: Message) => {
       message.content
     );
 
-    if (await multipleCountries(message.content)) {
+    if (matchedCountries.length > 1) {
       await message.reply(
         "Please only tell me 1 country for now, you can ask a member of the Support team about multiple nationalities :grin:"
       );
--- a/src/programs/where-are-you-from.ts
+++ b/src/programs/where-are-you-from.ts
@@ -20,6 +20,11 @@ const whereAreYouFrom = async (message: Message) => {
       message.content
     );
 
+    if (matchedCountries.length === 0) {
+      await message.reply("You must send the flag of the country you are in!");
+      return;
+    }
+
     if (matchedCountries.length > 1) {
       await message.reply(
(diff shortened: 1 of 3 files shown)


src/programs/where-are-you-from.ts
```diff
@@ -154,4 +154,16 @@ export const updateAfterRegionSelect = async (
...
+  if (
+    seperatedEmojis[0] + seperatedEmojis[1] !==
+      seperatedEmojis[2] + seperatedEmojis[3] &&
+    seperatedEmojis.length > 2
+  ) {
+    return true;
+  }
+  return false;
+};
+
```
geisterfurz007, line 168: The fix for detecting multiple countries should be moved to `CountryRoleFinder#getCountriesFromString` to make it only return the countries that are there.

Then we can have the tests in `__tests__/utils/country-role-finder.spec.ts` to verify everything works as expected (there is already one case commented out that relates to this issue).

src/programs/where-are-you-from.ts
```diff
@@ -154,4 +154,16 @@ export const updateAfterRegionSelect = async (
   }
 };
 
+const multipleCountries = async (message: string): Promise<boolean> => {
+  const seperatedEmojis = message.match(/\p{Emoji_Presentation}/gu);
+  if (
+    seperatedEmojis[0] + seperatedEmojis[1] !==
+      seperatedEmojis[2] + seperatedEmojis[3] &&
```
geisterfurz007, line 161: This allows 🇮🇳🇮🇳🇩🇪 to go through.

geisterfurz007 (changes requested):

src/programs/where-are-you-from.ts
hasanlatch007: I already said that this fix only detects for the first 2 flags, if they are the same everything is good to go, I didn't want to use a `for` to not have it use too many resources in case someone decided to spam flags, maybe I can limit the for to only check for the first 6 flags?

geisterfurz007: Oh, whoops, sorry about that!

I think it's fairly safe to use a loop; I have yet to see someone spam flags in a single message. Those with ill intent are more likely to spam large blocks of text or malicious links.

hasanlatch007:
This is still broken, but any ideas are always welcomed!

geisterfurz007 (approved):

PR #399, opened 2021-08-07
Feat: timeout with timer support

hasanlatch007:
Im still pretty new at the event system, so if you got any comments don't be afraid to be harsh :)

### DIFF
--- a/src/programs/timeout-timer.ts
+++ b/src/programs/timeout-timer.ts
@@ -17,7 +17,6 @@ const timeoutTimerIdentifier = "timeouttimer";
 
 interface TimeoutTimerData {
   userId: Snowflake;
-  channelId: Snowflake;
 }
 
 @Command({
@@ -28,33 +27,14 @@ interface TimeoutTimerData {
 })
 class TimeoutTimer implements CommandHandler<DiscordEvent.MESSAGE> {
   async handle(message: Message): Promise<void> {
-    if (message.system) {
-      await message.delete();
-      return;
-    }
-
     const words = message.content.split(/\s+/);
     const targetedUser = message.mentions.users.first();
     const time = Number(words[2]);
 
-    if (!targetedUser && !time) {
+    if (!targetedUser || !time) {
       return Tools.handleUserError(
         message,
         "The syntax for this command is `!timeoutTimer` {USER} {TIME IN MINUTES} {REASON (OPTIONAL)}"
-      );
-    }
-
-    if (!targetedUser) {
-      return Tools.handleUserError(
-        message,
-        "You have to ping the user you want to timeout!"
-      );
-    }
-
-    if (!time) {
-      return Tools.handleUserError(
-        message,
-        "You must set a time, it is interpreted in minutes"
       );
     }
 
@@ -78,7 +58,6 @@ class TimeoutTimer implements CommandHandler<DiscordEvent.MESSAGE> {
     executeTime.setMinutes(executeTime.getMinutes() + time);
     await TimerService.createTimer(timeoutTimerIdentifier, executeTime, {
       userId: targetedUser.id,
-      channelId: message.channel.id,
     });
   }
 }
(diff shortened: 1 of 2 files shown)


src/programs/timeout-timer.ts
```diff
@@ -0,0 +1,111 @@
...
+        "You have to ping the user you want to timeout!"
+      );
+    }
+
+    if (!time) {
+      return Tools.handleUserError(
+        message,
+        "You must set a time, it is interpreted in minutes"
+      );
+    }
```
geisterfurz007, line 59: I think we can shorten that by dropping the last two cases and using || instead of && in the first if.

src/programs/timeout-timer.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+@Command({
+  event: DiscordEvent.TIMER,
+  handlerIdentifier: timeoutTimerIdentifier,
+})
+class TimeoutPardon implements CommandHandler<DiscordEvent.TIMER> {
+  async handle(timer: Timer): Promise<void> {
+    const data = timer.data as unknown as TimeoutTimerData;
+    const channel = bot.channels.resolve(data.channelId) as TextChannel;
```
geisterfurz007, line 93: This channel is only used to access the guild. You can use `const guild = bot.guilds.resolve(process.env.GUILD_ID)` instead. That also means that you can drop the channelId from TimeoutTimerData.

src/programs/timeout-timer.ts
```diff
@@ -0,0 +1,111 @@
...
+    const memberGuild = channel.guild.members.resolve(data.userId);
```
geisterfurz007, line 94: I think the name `guildMember` fits better. The returned value is not a guild but a member. `GuildMember` is also the classname in Discord.js.

geisterfurz007 (changes requested):
Didn't try it out yet but the use of the events looks good!

src/programs/timeout-timer.ts
```diff
@@ -0,0 +1,81 @@
...
+  handlerIdentifier: timeoutTimerIdentifier,
+})
+class TimeoutPardon implements CommandHandler<DiscordEvent.TIMER> {
+  async handle(timer: Timer): Promise<void> {
+    const data = timer.data as unknown as TimeoutTimerData;
+    const guild = bot.guilds.resolve(process.env.GUILD_ID);
+    const guildMember = guild.members.resolve(data.userId);
+    const logChannel = guild.channels.cache.find(
+      (channel) => channel.name === "bot-output"
+    ) as TextChannel;
```
geisterfurz007, line 69: Missed that in my last review; you can use `textLog` from `src/common/moderator.ts` to log to bot-output instead of looking up the channel here, so this could be removed (and the two `logChannel.send` calls replaced with textLog).

geisterfurz007 (changes requested):
Rest looks and works fantastically!

src/programs/timeout-timer.ts
hasanlatch007: Also changed it in `timeout.ts`

src/programs/timeout-timer.ts
```diff
@@ -0,0 +1,78 @@
...
+      return Tools.handleUserError(
+        message,
+        "The syntax for this command is `!timeoutTimer` {USER} {TIME IN MINUTES} {REASON (OPTIONAL)}"
+      );
+    }
+
+    const targetedGuildMember = message.guild.members.resolve(targetedUser);
+    const timeoutRole = message.guild.roles.cache.find(
+      (r) => r.name === "Time Out"
+    );
```
geisterfurz007, line 37:
```suggestion
    const timeoutRole = Tools.getRoleByName("Time Out", message.guild);
```

src/programs/timeout-timer.ts
```diff
@@ -0,0 +1,78 @@
...
+@Command({
+  event: DiscordEvent.TIMER,
+  handlerIdentifier: timeoutTimerIdentifier,
+})
+class TimeoutPardon implements CommandHandler<DiscordEvent.TIMER> {
+  async handle(timer: Timer): Promise<void> {
+    const data = timer.data as unknown as TimeoutTimerData;
+    const guild = bot.guilds.resolve(process.env.GUILD_ID);
+    const guildMember = guild.members.resolve(data.userId);
+    const timeoutRole = guild.roles.cache.find((r) => r.name === "Time Out");
```
geisterfurz007, line 67:
```suggestion
    const timeoutRole = Tools.getRoleByName("Time Out", guild);
```

geisterfurz007 (changes requested):
Last one, I promise!

geisterfurz007 (approved):

PR #400, opened 2021-08-08
Chore: Improve Moderation.ts

hasanlatch007:
To clean up our files a bit better for the re-write, `moderation.ts` could add functions like getRole to resolve roles, resolve guildMembers etc... If this is completely useless let me know :)

### DIFF
--- a/src/common/moderator.ts
+++ b/src/common/moderator.ts
@@ -3,6 +3,7 @@ import {
   GuildMember,
   Message,
   PartialGuildMember,
+  Role,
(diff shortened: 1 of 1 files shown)


geisterfurz007:
This appears to be a duplication of `Tools#getRoleByName`:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/1730d1d80534c8ad73977647d7ddf1706c0d1aca/src/common/tools.ts#L79-L83

PR #401, opened 2021-08-08
feat: subtriggers in the event and tree utility

geisterfurz007:
Commands like !voice allow for a variety of sub commands like create, limit, etc. To more easily support splitting the code for those, this PR adds a subtrigger property to the message event handler options.

Now instead of having one handler for !voice which has to parse the sub command and call functions accordingly, you can now create a command with trigger !voice and subTrigger create to have a class only to handle creating a voice room.

To not go deeper down the hand creating tree rabbit hole for the message handler, this PR also adds a utility `addToTree` that adds a handler with a list of keys recursively creating new layers to match the length of the key array. Existing code that adds handler to the tree has been refactored to use that function.


src/event-distribution/events/events.ts
```diff
@@ -49,7 +49,7 @@ import {
   TimerEventHandlerOptions,
   TimerHandlerFunction,
 } from "./timer";
-import { Prisma, Timer } from "@yes-theory-fam/database/client";
+import { Timer } from "@yes-theory-fam/database/client";
```
adrian-goe, line 52: Is it correct, that this is removed?

hasanlatch007: I dont see prisma used in that file in general also isn't the correct import for prisma is `import prisma from ../prisma`?

geisterfurz007: Yes, this was correctly removed. It was a relic from the timer events, where I originally used `Prisma.JsonValue` as one of the parameters to the handler before switching to directly passing the `Timer` objects themselves. The unused import was never removed.

Re @hasanlatch007: For access to the database, the import you mentioned, would be the correct one; the one used there was to get access to one the various types/interfaces (namely `JsonValue`) that are exported under the `Prisma` namespace from the generated client.

PR #412, opened 2021-08-14
chore: update contributing to mention db migration

geisterfurz007:
The setup guide for YesBot didn't mention migrating or pushing the schema to the DB yet.

This PR includes an additional step in the contributing guide to mention that required command. It also fixes a small left-over phrase that wasn't accurate anymore.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -151,7 +151,13 @@ yarn install
 
 3. Create an `.env` in the root directory and copy the `.env.example`. Change the values to match your test server.
 
-4. Start the bot
+4. Set up the tables in the database
+
+```bash
+yarn prisma db push
+```
+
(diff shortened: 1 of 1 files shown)


hasanlatch007:
When going through the installation doesn't it automatically generate the schema tables for the db? Never remember using `db:push` when installing the bot

geisterfurz007:
> When going through the installation doesn't it automatically generate the schema tables for the db? Never remember using `db:push` when installing the bot

It doesn't. I am not sure anymore when you joined but TypeORM did that automatically. `yarn prisma db push` is similar to `yarn prisma migrate dev` just that it ignores the migration column. Instead it just creates the tables and columns as defined in the schema which should be enough for getting started with the bot.

hasanlatch007 (approved):

PR #418, opened 2021-08-20
Chore: Rewrite polls

hasanlatch007:
I have rewritten `polls.ts` with the new event handlers but there was a function exported to the old event handler `reaction_add.ts` not sure what it really does or if I did it correctly since i left the emoji string empty.

### DIFF
--- a/src/programs/polls.ts
+++ b/src/programs/polls.ts
@@ -119,17 +119,15 @@ class Polls implements CommandHandler<DiscordEvent.MESSAGE> {
 @Command({
   event: DiscordEvent.REACTION_ADD,
   channelNames: ["polls"],
-  description: "This... I actually have no idea",
+  requiredRoles: ["Support"],
   emoji: "",
+  description:
+    "This handler mirrors a Support member's new reaction on a poll.",
 })
 class ModeratorPollMirror implements CommandHandler<DiscordEvent.REACTION_ADD> {
   async handle(reaction: MessageReaction, user: User | PartialUser) {
     const { message } = reaction;
-    const { channel } = message;
-    if (channel instanceof DMChannel || channel.name !== "polls") return;
-    const member = channel.guild.member(user.id);
 
-    if (!hasRole(member, "Support")) return;
     const fetched = await reaction.users.fetch();
     if (fetched.size > 1) return;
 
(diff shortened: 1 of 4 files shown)


src/programs/polls.ts
```diff
@@ -78,51 +79,61 @@ const partition = <T>(items: T[], size: number): T[][] => {
...
+        await reactMessage.react(emoji);
+      }
+    }
+  }
+}
+
+@Command({
+  event: DiscordEvent.REACTION_ADD,
+  channelNames: ["polls"],
+  description: "This... I actually have no idea",
```
geisterfurz007, line 122: Mirrors a Support's reaction to a poll so it doesn't look like the support actually voted.

This was much more relevant when people would add a C to their polls while the bot couldn't pick up most reactions itself. Essentially what happens is that the bot adds the reaction if it's the first and added by a Support (since that is understood as the Support wanting to add that option to a poll) and then removes the Support's reaction again to only leave YesBot's one behind.

So for the description, something like `Mirrors a Support member's new reaction on a poll.`, or similar.

src/programs/polls.ts
```diff
@@ -78,51 +79,61 @@ const partition = <T>(items: T[], size: number): T[][] => {
...
+  emoji: "",
+})
+class ModeratorPollMirror implements CommandHandler<DiscordEvent.REACTION_ADD> {
+  async handle(reaction: MessageReaction, user: User | PartialUser) {
+    const { message } = reaction;
+    const { channel } = message;
+    if (channel instanceof DMChannel || channel.name !== "polls") return;
```
geisterfurz007, line 129: This line can be removed; it will be guaranteed to be in polls and not in DMs.

src/programs/polls.ts
```diff
@@ -78,51 +79,61 @@ const partition = <T>(items: T[], size: number): T[][] => {
...
+    const member = channel.guild.member(user.id);
+
+    if (!hasRole(member, "Support")) return;
```
geisterfurz007, line 132: You can specify `requiredRoles: ["Support"]` in the `Command`'s options. That also removes the need for the `member` variable which in turn removes the need for the `channel` variable; both can be removed.

geisterfurz007 (changes requested):

geisterfurz007 (approved):
Noice!

PR #442, opened 2021-08-21
Chore: Refactored nitro-colors

### LINKED ISSUE #439: Convert Nitro Role handling to new event system
Parts of this were stuffed into src/events/guild-member-update.ts iirc.

hasanlatch007:
!closes #439

### DIFF
--- a/src/events/reaction-add.ts
+++ b/src/events/reaction-add.ts
@@ -3,6 +3,10 @@ import Tools from "../common/tools";
 import { GroupManagerTools } from "../programs";
 import { hasRole } from "../common/moderator";
 import prisma from "../prisma";
+import {
+  isColorSelectionMessage,
+  memberHasNitroColor,
+} from "../programs/nitro-colors";
 
 const reactionAdd = async (
   messageReaction: MessageReaction,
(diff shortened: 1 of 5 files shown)


src/events/reaction-add.ts
```diff
@@ -37,23 +37,7 @@ const addRolesFromReaction = async (
...
-          dm.send(
-            "You can't assign yourself a new colour yet, please wait until the end of the month!"
-          )
-        );
-
-      await messageReaction.users.remove(guildMember);
-    } else {
-      await guildMember.roles.add(roleToAdd);
-    }
+    await guildMember.roles.add(roleToAdd);
```
geisterfurz007, line 40: Having this around breaks things. The reactions don't honor the exception for nitro anymore so what happens now is that the role is added, then the handler is called and detects that the user already has a nitro role (which was added here) and removes the reaction again.

With this left over, we also allow having multiple color roles. I am not entirely sure how that works because removing the reaction should also remove the role but that doesn't appear to happen in this case.

geisterfurz007 (changes requested):

src/events/reaction-add.ts
hasanlatch007: Not really sure why the role was not being removed but figured it was mostly both events colliding with each other I'm mostly assuming that the old event was quicker and added the role immediately while the event in `nitro-colors.ts` was still running it sees that they already have a role and only removes their reaction and not the role. Anyways since I don't feel like deleting half of `reaction-add.ts` YET, decided to add a simple `if` and let the event in `nitro-colors.ts` handle adding the roles or removing them.

src/events/reaction-add.ts
```diff
@@ -33,27 +37,15 @@ const addRolesFromReaction = async (
     },
   });
 
+  const guildMember =
+    guild.member(user.id) ?? (await guild.members.fetch(user.id));
+  //Since most of the bot isn't refactored yet, this must stay for the old and new event do not collide together. Color Roles are handled in Nitro-colors.ts
+  if (isColorSelectionMessage(messageId) && memberHasNitroColor(guildMember))
```
geisterfurz007, line 43: This still suffered from some weird race-condition. This should fix it though (we don't want this file to handle nitro colors, no matter whether the member already has a nitro color or not):
```suggestion
  if (isColorSelectionMessage(messageId))
```

geisterfurz007 (changes requested):

geisterfurz007 (approved):

PR #445, opened 2021-08-21
fix: require only one role in requiredRoles

geisterfurz007:
Requiring all roles in the requiredRoles array to be present was a bad idea; requiring only one makes so much more sense in our usecases.

This PR renames requiredRoles to allowedRoles and fixes the issue described above.


hasanlatch007 (approved):
LGTM!

PR #447, opened 2021-08-22
fix: allowed-roles

geisterfurz007:
Array.some returns false if the array is of zero length. In our usecase an empty array should indicate that everyone may use it, however. This PR fixes the behaviour of allowedRoles.


hasanlatch007 (approved):
LGTM!

PR #448, opened 2021-08-22
fix: move to google sheets API v4

geisterfurz007:
Google Sheets API v3 which YesBot is currently using, has been shut down on August 2nd, causing `@someone` and `!topic` in philosophy to fail.

This PR introduces a common method to use to access the first column of a sheet when provided the sheet's ID which covers both usecases mentioned above.


hasanlatch007 (approved):
LGTM!

PR #455, opened 2021-08-24
Chore: Cleaned up code and moved function to the right files

hasanlatch007:


geisterfurz007:
Don't duplicate code. If you want to get rid of custom-methods.ts (which I heavily support!), move it to src/common/tools.ts for example. If we ever need to change something in the communication with the sheets, we will have to do it twice like this and doing work twice isn't cool :c

geisterfurz007 (approved):

PR #465, opened 2021-08-29
Chore: Refactor !voice

### LINKED ISSUE #433: Convert !voice to new event system
As this is a larger command, split the functionality in separate files in src/programs/voice. Common logic can be put in src/programs/voice/common.ts.

Currently blocked by #401.

hasanlatch007:
This still needs testing but I wanted to leave this PR here to show the progress so far.
Early reviews are always welcomed (Michel 👀)
!closes #433


hasanlatch007:
My internet kept crashing during testing, which made me lost track of what I tested. I heavily tested the implemented `TIMER EVENT` to be able to delete the channel or requestNewHost and it works like a charm!

hasanlatch007:
This is not the cleanest approach for splitting up the functions, but it is better than exporting a bunch of constants from `tools.ts` for variables like `maxLimit` and the identifiers

geisterfurz007 (changes requested):

src/programs/voice/voice-timer.ts
```diff
@@ -0,0 +1,264 @@
...
+
+@Command({
+  event: DiscordEvent.VOICE_STATE_UPDATE,
+  changes: [VoiceStateChange.LEFT, VoiceStateChange.SWITCHED_CHANNEL],
+})
+class RequestNewHostIfNeeded
+  implements CommandHandler<DiscordEvent.VOICE_STATE_UPDATE>
+{
+  async handle(oldState: VoiceState, newState: VoiceState): Promise<void> {
+    if (!oldState.channel || oldState.channelID === newState.channelID) return;
```
geisterfurz007, line 120: `oldState.channel` should never be falsy so the first part of the condition can be removed.

src/programs/voice/voice-timer.ts
```diff
@@ -0,0 +1,264 @@
...
+    await requestOwnershipTransfer(channel, mapping);
+  }
+}
+
+@Command({
+  event: DiscordEvent.VOICE_STATE_UPDATE,
+  changes: [VoiceStateChange.JOINED, VoiceStateChange.SWITCHED_CHANNEL],
+})
+class VoiceOnDemandPermissions
+  implements CommandHandler<DiscordEvent.VOICE_STATE_UPDATE>
```
geisterfurz007, line 87: This is not timer related and not a good fit for this file; maybe voice-create is more suitable since it's somewhat tied to that?

src/programs/voice/voice-create.ts
```diff
@@ -0,0 +1,122 @@
...
+import {
+  Command,
+  CommandHandler,
+  DiscordEvent,
+} from "../../event-distribution";
+import { TimerService } from "../timer/timer.service";
+import VoiceOnDemandTools from "./common";
+
+const emojiPool = ["📹", "💬", "📺", "🎲", "🎵", "🏋️"];
+const voiceOnDemandDeleteIdentifier = "voiceondemandchanneldelete";
```
geisterfurz007, line 13: This is duplicated from voice-timer.ts; it's probably best to export the identifier from there and use it here to avoid things breaking if the identifier ever changes and is only changed in one place.

src/programs/voice/voice-timer.ts
```diff
@@ -0,0 +1,264 @@
...
+      await channel.delete();
+      await removeMapping(channel.id);
+      return;
+    }
+    //This is necessary to loop the timer until the channel is deleted
+    const executeTime = new Date();
+    executeTime.setMinutes(executeTime.getMinutes() + 1);
+    await TimerService.createTimer(voiceOnDemandDeleteIdentifier, executeTime, {
+      channelId: channel.id,
+    });
```
geisterfurz007, line 58: Is it? The check shouldn't be done every minute; it should be triggered by the last person leaving.

geisterfurz007 (changes requested):
The bot is currently missing an implementation to check all channels on startup as is currently done:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/88e5be34f05abf01e5fdac1cbc83f7e163dd3c60/src/programs/voice-on-demand.ts#L439-L444

src/programs/voice/voice-timer.ts
geisterfurz007: Some further reasoning: This could introduce flawed behaviour. Imagine the timer triggers every full minute. Last member leaves at 13:00:58; the channel closes 2 seconds after the last person left instead of a minute. Thus checking periodically is not ideal.

src/programs/voice/common.ts
```diff
@@ -0,0 +1,72 @@
...
+export const defaultLimit = 5;
+
+class VoiceOnDemandTools {
+  static stringToWords(inputStr: string): Array<string> {
+    return <string[]>inputStr.split(" ");
+  }
+
+  static async handleLimitCommand(
+    message: Message,
+    requestedLimit: any,
```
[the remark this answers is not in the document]
hasanlatch007: It would break the `Number()` function when setting it as a string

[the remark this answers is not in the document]
geisterfurz007: In TypeScript, declaring a parameter as `any` tells the compiler to just shut up which is the reason it should be avoided. The errors you get when declaring requestedLimit as string are correct, because you are not supposed to pass a string into Math.floor or Math.min.

Here is what I would propose:

```ts
  static async handleLimitCommand(
    message: Message,
    requestedLimitString: string, // Renamed this variable and changed type
    createCommand?: boolean
  ): Promise<number> {
    const requestedLimit = Number(requestedLimitString); // Added this line

    if (!requestedLimit && createCommand) return defaultLimit;
    
    /* Rest of the function */
```

geisterfurz007 (commented):
Sadly the timers break one thing and quite subtly at that.

1) Create channel
2) Connect after 30 seconds
3) Disconnect after 5 seconds

-> Expectation: Channel is deleted 1 minute after disconnecting
-> Actual: Channel is deleted 25 seconds after disconnecting because the previous timer is not cancelled

I am afraid this might end up blocked for a bit while I add stuff to cancel timers to the timer-service.

geisterfurz007:
Scrolling through open PRs, I found this is not blocked anymore, now that timer cancels are a thing.

hasanlatch007:
> Scrolling through open PRs, I found this is not blocked anymore, now that timer cancels are a thing.

Yes I know, but I left it mostly blocked until my laptop comes back from repairs `voice` requires lots of testing before even submitting new code for it, with this slow mac that is near impossible :'), I'll start working on them once I have the chance.

hasanlatch007:
I'll rework this on a new PR since this one is too old.

PR #466, opened 2021-08-30
chore: Refactored !role

### LINKED ISSUE #436: Convert !role to new event system
As this is a larger command, split the functionality in separate files in src/programs/role. Common logic can be put in src/programs/role/common.ts.

Currently blocked by #401.

hasanlatch007:
Took a different approach for `!role add` instead of having to copy the channel id and messageId you can use the same tactic of `!meta` which is simply replying to the message, everything will be deleted and all messages send relating to the command will be deleted leaving no trace behind and making the Support life easier instead of having to copy so much.
!closes #436


src/programs/role/react-role-delete.ts
```diff
@@ -0,0 +1,62 @@
...
+@Command({
+  event: DiscordEvent.MESSAGE,
+  trigger: "!role",
+  subTrigger: "delete",
+  allowedRoles: ["Support"],
+  description: "This",
+})
+class DeleteReactRoleObjects implements CommandHandler<DiscordEvent.MESSAGE> {
+  async handle(message: Message): Promise<void> {
+    console.log(message.content.split(" "));
```
geisterfurz007, line 22: Debugging output spotted 👁️👄👁️

geisterfurz007 (changes requested):

src/programs/role/react-role-delete.ts
hasanlatch007: :')

geisterfurz007 (approved):
LGTM!

PR #468, opened 2021-08-31
Chore: Refactored ticket.ts

### LINKED ISSUE #429: Rewrite the ticket system with the new event system
The ticket system is an absolute mess atm. It should be rebuilt almost from scratch to be more extensible in the future and support the new event system.

hasanlatch007:
I rewrote anything related to the discord commands but left the other functions as they are.
!closes #429


src/programs/ticket.ts
```diff
@@ -67,58 +53,110 @@ const ticket = async (message: Message, type: string) => {
...
+      channelName,
+      channelOptions
+    );
+
+    await ticketChannel.send(
+      `Hi ${message.member.toString()}, please list the details of your shoutout below. A ${moderatorRole.toString()} will be with you as soon as possible.`
+    );
+  }
+}
+//Commented out code yes i know but fiyesta's are discontinued due to the pandemic.
```
geisterfurz007, line 67: I don't mind the commented code per se but I am afraid that once they are reopened, this block will just be uncommented with all the duplicate code that's hidden inside of it.

There aren't many differences between the ticket types. Originally the channels were supposed to be created in different channels (fiyestas for fiyestas and mod section for shoutouts) and handled by different moderator roles (engineer for fiyestas and coordinators for shoutouts). Through the new server staff structure, this doesn't work because the equivalent to both roles originally responsible for their tickets are not tied to the Support role anymore.

The only differences I can recall are the names of the channel and the greeting message. Maybe we can move everything else to some common function?

geisterfurz007 (changes requested):

geisterfurz007 (changes requested):

geisterfurz007 (approved):
Passed!

PR #480, opened 2021-09-07
Chore: Re-added optional options to !role add

### LINKED ISSUE #479: Add optional options to !role for messageId and channelId
Requested by the man himself josh to use both reply function and the ID options for some cases to avoid soft pings

hasanlatch007:
The only way I thought of approaching this without having multiple `if` inside each other.
My creativity is sometimes frightening 
!closes #479

### DIFF
--- a/src/programs/role/react-role-add.ts
+++ b/src/programs/role/react-role-add.ts
@@ -1,4 +1,4 @@
-import { Message, MessageEmbed, TextChannel } from "discord.js";
+import { Channel, Message, MessageEmbed, TextChannel } from "discord.js";
 import Tools from "../../common/tools";
 import { textLog } from "../../common/moderator";
 import prisma from "../../prisma";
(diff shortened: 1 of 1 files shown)


src/programs/role/react-role-add.ts
```diff
@@ -17,62 +17,102 @@ const logger = createYesBotLogger("programs", "role-add");
...
+
+    let [, , reaction, roleId, messageId, channelId] =
+      message.content.split(" ");
+
+    if (!channelId || !messageId) {
+      hasChannelId = false;
+      hasMessageId = false;
+    }
+
+    if (message.reference && !hasChannelId && !hasMessageId) {
```
geisterfurz007, line 35: I would just ignore everything if there was a message reference, tbh. Flow could be:

```ts
const [/*stuff*/] = message.content.split(" ");
if (message.reference) { /* Use channel and message from the reference + return */ }
if (/* any parameters missing */) { /* something something userError + return */ }
/* use the ids from stuff */
```

Would reduce the amount of variables you have to track while reading the code as `hasChannelId` and `hasMessageId` aren't needed anymore. You also get away with one less `if` ;)

src/programs/role/react-role-add.ts
```diff
@@ -17,62 +17,102 @@ const logger = createYesBotLogger("programs", "role-add");
...
+  description: "This handler is to add a reaction that can give a role",
 })
 class AddReactRoleObject implements CommandHandler<DiscordEvent.MESSAGE> {
   async handle(message: Message): Promise<void> {
-    if (!message.reference) {
-      await Tools.handleUserError(
+    let hasChannelId = true;
+    let hasMessageId = true;
+
+    let [, , reaction, roleId, messageId, channelId] =
```
geisterfurz007, line 27: Prefer `const`

geisterfurz007 (changes requested):

geisterfurz007 (approved):

PR #481, opened 2021-09-09
refactored map.ts

### LINKED ISSUE #427: Convert !map and !mapadd to new event system

AdamWarden:
!closes #427

### DIFF
--- a/src/programs/map.ts
+++ b/src/programs/map.ts
@@ -7,29 +7,31 @@ import { CountryRoleFinder } from "../utils/country-role-finder";
   event: DiscordEvent.MESSAGE,
   trigger: "!map",
   channelNames: ["bot-commands"],
-  description: "map to new event system",
+  description: "command to view the yf map",
+})
+@Command({
+  event: DiscordEvent.MESSAGE,
+  trigger: "!mapadd",
+  channelNames: ["bot-commands"],
+  description: "adds location of user to yf map",
 })
 class ShowMap implements CommandHandler<DiscordEvent.MESSAGE> {
   async handle(message: Message): Promise<void> {
-    const map = async (message: Message) => {
-      await message.reply(
-        "you can find the map here: " +
-          process.env.MAP_LINK +
-          "\nIf you want to be added to it, type !mapadd [city, country]"
-      );
-    };
+    await message.reply(
+      "you can find the map here: " +
+        process.env.MAP_LINK +
+        "\nIf you want to be added to it, type !mapadd [city, country]"
+    );
 
-    const mapAdd = async (message: Message) => {
-      const split = message.content.split(" ");
-      split.shift();
-      const city = split.join(" ");
-      if (!city) {
-        await Tools.handleUserError(
-          message,
-          "You need to add the city you are from!"
-        );
-        return;
-      }
+    const split = message.content.split(" ");
+    split.shift();
+    const city = split.join(" ");
+    if (!city) {
+      await Tools.handleUserError(
+        message,
+        "You need to add the city you are from!"
+      );
+      return;
 
       const countries = message.member.roles.cache
         .filter((role) => CountryRoleFinder.isCountryRole(role.name))
@@ -50,6 +52,6 @@ class ShowMap implements CommandHandler<DiscordEvent.MESSAGE> {
       await message.reply(
         "I messaged the maintainer of the map, they will add you to it soon!"
       );
-    };
+    }
   }
 }
--- a/src/programs/map.ts
+++ b/src/programs/map.ts
@@ -1,44 +1,55 @@
 import { Message } from "discord.js";
 import Tools from "../common/tools";
+import { Command, CommandHandler, DiscordEvent } from "../event-distribution";
 import { CountryRoleFinder } from "../utils/country-role-finder";
 
-export const map = async (message: Message) => {
-  await message.reply(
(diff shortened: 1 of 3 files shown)


src/programs/map.ts
```diff
@@ -1,44 +1,55 @@
...
+@Command({
+  event: DiscordEvent.MESSAGE,
+  trigger: "!map",
+  channelNames: ["bot-commands"],
+  description: "map to new event system",
+})
+class ShowMap implements CommandHandler<DiscordEvent.MESSAGE> {
+  async handle(message: Message): Promise<void> {
+    const map = async (message: Message) => {
+      await message.reply(
```
hasanlatch007, line 15: You no longer need to have `map` as a function.
In TS when you use the `=>` or called "Fat arrow" is a way to write a concise function, with the new event handlers this technique is no longer since the class object is now our function, plus `async handle()` is our actual function thus you can delete that line.

src/programs/map.ts
```diff
@@ -1,44 +1,55 @@
...
-  if (!city) {
-    await Tools.handleUserError(
-      message,
-      "You need to add the city you are from!"
-    );
-    return;
-  }
+    const mapAdd = async (message: Message) => {
+      const split = message.content.split(" ");
+      split.shift();
```
hasanlatch007, line 24: See comment above

src/programs/map.ts
```diff
@@ -1,44 +1,55 @@
...
-      process.env.MAP_LINK +
-      "\nIf you want to be added to it, type !mapadd [city, country]"
-  );
-};
+@Command({
+  event: DiscordEvent.MESSAGE,
+  trigger: "!map",
+  channelNames: ["bot-commands"],
+  description: "map to new event system",
+})
```
hasanlatch007, line 11: This is currently only one command which is triggered by the trigger set "!map", which is different from "!mapadd" so you're going to have to split them up.
Also, the description is preferably to explain the command and not the code.

hasanlatch007 (changes requested):
There are also leftovers from map.ts that you're going to have to delete in these files
https://github.com/Yes-Theory-Fam/yesbot-ts/blob/master/src/events/message.ts
https://github.com/Yes-Theory-Fam/yesbot-ts/blob/master/src/programs/index.ts

src/programs/map.ts
```diff
@@ -1,44 +1,57 @@
...
-  const city = split.join(" ");
-  if (!city) {
-    await Tools.handleUserError(
-      message,
-      "You need to add the city you are from!"
+@Command({
+  event: DiscordEvent.MESSAGE,
+  trigger: "!map",
+  channelNames: ["bot-commands"],
+  description: "command to view the yf map",
```
hasanlatch007, line 10:
```suggestion
  description: "This handler is to show the Yes Theory Fam map",
```

src/programs/map.ts
```diff
@@ -1,44 +1,57 @@
...
+})
+@Command({
+  event: DiscordEvent.MESSAGE,
+  trigger: "!mapadd",
+  channelNames: ["bot-commands"],
+  description: "adds location of user to yf map",
```
hasanlatch007, line 16:
```suggestion
  description: "This handler is to request to add your location to the Yes Theory Fam map",
```

src/programs/map.ts
```diff
@@ -1,44 +1,57 @@
...
+      City/Location: ${city}
+      Countries: ${countries.join(", ")}
+      Link to the message calling the command: ${message.url}`;
+      await maintainerDm.send(infoString);
+      await message.reply(
+        "I messaged the maintainer of the map, they will add you to it soon!"
+      );
+    }
+  }
+}
```
hasanlatch007, line 57: This is an unreachable code, you added the code after the return. You have to take out the code after your return and put it outside the if brackets
```suggestion
    if (!city) {
      await Tools.handleUserError(
        message,
        "You need to add the city you are from!"
      );
      return;
      }
      //rest of code here
```

src/programs/map.ts
```diff
@@ -1,44 +1,57 @@
...
-
-export const mapAdd = async (message: Message) => {
-  const split = message.content.split(" ");
-  split.shift();
-  const city = split.join(" ");
-  if (!city) {
-    await Tools.handleUserError(
-      message,
-      "You need to add the city you are from!"
+@Command({
```
hasanlatch007, line 6: You have to put the code that is related to `!mapadd` to this just imagine 2 separate commands with their own code. You can't stack them like this.

src/programs/map.ts
```diff
@@ -1,44 +1,57 @@
...
+      const countries = message.member.roles.cache
+        .filter((role) => CountryRoleFinder.isCountryRole(role.name))
+        .map((role) => CountryRoleFinder.getCountryByRole(role.name));
+
+      const maintainerDm = await message.guild.members
+        .resolve(process.env.MAP_ADD_DM_USER_ID)
+        .user.createDM();
+      const author = message.member;
+      const infoString = `Someone new wants to be added to the map!
+      UserID: ${message.author.id}
```
hasanlatch007, line 45:
```suggestion
      UserID: ${author.id}
```

hasanlatch007 (changes requested):

geisterfurz007:
This hasn't seen activity in several months; closing. Feel free to let me know if you pick this up again!

PR #495, opened 2021-09-19
feat: upgrade !travel

### LINKED ISSUE #489: Create a new form of ticket for the traveling-together channel
Let's be honest: #traveling-together is a mess at the moment; it was suggested that we use the recently refactored ticket system to have travel request sent in a more streamlined manner and to declutter the channel which is currently full of both people's requests *and* the Support's replication of it.

That should help reduce the amount of chaos in the channel and help Supports to keep track of what already has a thread and what doesn't until a neater solution is finally ready for launch.

### Update

After a recent mod meeting, this has been extended:

- Bot command `!travel` that collects the data for the current "form"
    - Requires "Seek Discomfort" (sends error message when missing role)
    - Puts it into the format currently used in traveling-together
    - Puts it up for Supports to approve it (Member shall have patience)
    - If approved sends in traveling-together (Support creates a thread until YesBot is moved to Discord.js v13)
    - If declined:
        - Message is sent to ticket channel
        - Member has to redo it again with the changes asked by the Support

geisterfurz007:
This PR aims to provide a new command to streamline the process for creating new posts in traveling together.

Closes #489, see that issue for more information.


src/programs/tickets/travel/open-travel-ticket.ts
```diff
@@ -0,0 +1,44 @@
...
+@Command({
+  event: DiscordEvent.MESSAGE,
+  trigger: "!travel",
+  description: "This handler is to create a travel ticket.",
+  stateful: false,
+})
+class OpenTravelTicket implements CommandHandler<DiscordEvent.MESSAGE> {
+  async handle(message: Message): Promise<void> {
+    // It's not possible yet to react to missing permissions. See #496
+    if (!message.member.roles.cache.some((r) => r.name === "Seek Discomfort")) {
```
hasanlatch007, line 21: This can be replaced with `hasRole`
(Not necessary but makes the code better to look at :)

hasanlatch007 (changes requested):
The code runs perfectly and I didn't see any bugs or found a way to break anything but there are general inputs I would like to add
Since this is now following a form request, it would be better if the user would have a confirmation message with all the information he added before sending it for review.
If in case he messed up the date for example.
Also if possible add the optional object of who is traveling in case it was multiple users from the server 
Ask if multiple are traveling and they can input their names or discord tag
And last note I want to add is that when force closing or closing the ticket it would log having a double copy from the `travel-approval` nevertheless if the messages were to be edited before closing the ticket it would not log properly what the user send.
Thus having 2 different copies of logs
Could be fixed when creating the channel not giving the user the ability to edit messages.

geisterfurz007:
> Since this is now following a form request, it would be better if the user would have a confirmation message with all the information he added before sending it for review.

Good point, done!

> Also if possible add the optional object of who is traveling in case it was multiple users from the server

Done (it's a bit rough to use but should work for now 👍)!

> And last note I want to add is that when force closing or closing the ticket if the messages were to be edited before closing the ticket it would not log properly what the user send. Thus having an incorrect log Could be fixed when creating the channel not giving the user the ability to edit messages.

I didn't understand this but I don't think it is going to be an issue. Also you cannot revoke someone's permission to edit messages, afaik.

geisterfurz007:
> The first-mentioned user is the user who was tagged as the travel buddy and not the one who created the ticket see below the format of the ticket

Good catch, fixed @hasanlatch007!

PR #500, opened 2021-09-22
Feat: Removed old events for reactions and added !channelToggle remove

### LINKED ISSUE #498: Allow deletion of channel-toggles
`!group toggle ...` allows creating a reaction that gives per-user permissions to a certain channel.

There is nothing to remove those, meaning a Support has to manually remove all reactions and a Developer has to remove the respective entry from the database which is quite tedious.

Granted, this doesn't happen often but is inconvenient when it does.

This might also be a good time to break this thing out of the `!group` command because it has nothing to do with the pingable groups. Maybe `!channelToggle add` / `!channelToggle remove` might work (other ideas most certainly welcome)?

hasanlatch007:
Since `role-react` and `!group-toggle` are being refactored any events related to reactions in the old event folder can be deleted
This still needs testing and reformatting :)

!closes #498


hasanlatch007:
Github why are you being shit?

hasanlatch007:
This PR adds multiple fixes for `reactionRole`, where when adding the reaction the bot would also get the role.

src/programs/channel-toggle/channel-toggle-add.ts
```diff
@@ -0,0 +1,85 @@
...
+import {
+  Command,
+  CommandHandler,
+  DiscordEvent,
+} from "../../event-distribution";
+import { createYesBotLogger } from "../../log";
+import prisma from "../../prisma";
+import { backfillReactions, getOrCreateMessage } from "./common";
+
+const logger = createYesBotLogger("program,", "channelToggleRemove");
```
geisterfurz007, line 12:
```suggestion
const logger = createYesBotLogger("program", "channelToggleRemove");
```
;)

geisterfurz007 (changes requested):

geisterfurz007:
Rest LGTM, nice work!

geisterfurz007 (approved):

PR #514, opened 2021-09-30
feat(currency): track activity

adrian-goe:
track server activity time for users

depends on
https://github.com/Yes-Theory-Fam/database/pull/66

PR #515, opened 2021-09-30
Introduce currency

adrian-goe:
activity will be converted into currency

depends on https://github.com/Yes-Theory-Fam/database/pull/67 and https://github.com/Yes-Theory-Fam/yesbot-ts/pull/514


hasanlatch007 (changes requested):
`Package.json` does not include the new module used `node-cron`

hasanlatch007 (approved):
LGTM!

PR #560, opened 2021-10-23
Upgrade @types/node

geisterfurz007:
Since Renovate didn't pick up on the major node upgrade in the types package, this PR takes care of that.


geisterfurz007:
Superseeded by #576

PR #582, opened 2021-11-04
Feat: Expand menu.ts and let bot handle break role

### LINKED ISSUE #196: Let the bot handle the Break role
Break was solely built as voluntarily picked role so there is no need to worry about potential abuse by people who are timed out. This issue *only* talks about Break, *not* Time Out for that reason.

More and more people are taking up on the offer of getting the Break role assigned, removing access to most channels. They later contact moderators to get it removed.

## Idea

Implement a `!break` command that works everywhere. It prompts the user whether they are sure they want the Break role and when confirmed assigns the tag. When the user wants it removed, they can send YesBot a DM and an additional option is included in the DM menu that, when clicked, removes the Break role.

## Usage

```
!break
!menu # in DMs with YesBot, then click the respective icon
```

## Blocked

In an ongoing issue, the handling of DMs is being slightly reworked. Work on this issue should not start until that is merged. This issue will be updated once that happened.

## Thoughts

 - Is there any reason to save these things in a database? The only thing that comes to mind is the 24 hour limit which was set to prevent people from asking for a Break for 20 minutes. If we can expect sensible use of the command, we can get away with not saving the time. If we can't, we have to to prevent API restrictions because of the flood of channel permission changes the role causes.
 - If we *do* need a database, it might be worth thinking about adding mod only variants of the command `!break @user` and `!break remove @user` to properly modify the database when handing out the role.

hasanlatch007:
I split the commands a bit so when @adrian-goe adds the coin feature he would be easily able to implement it.
Currently, this only sets the preparation for the new feature of letting the bot handle the `break` role
Also added it to its own personal folder for future usage for Adrian's features :)

!closes #196


adrian-goe:
That's 2 seperate things in one MR.
Also, you need to rebase.

hasanlatch007:
This requires https://github.com/Yes-Theory-Fam/database/pull/95 to be merged.

src/programs/menu/break-handler/addBreak.ts
```diff
@@ -0,0 +1,44 @@
+import prisma from "../../../prisma";
```
geisterfurz007, line 1: File name should be in line with our naming convention -> `add-break.ts`.

src/programs/menu/break-handler/addBreak.ts
```diff
@@ -0,0 +1,44 @@
...
+import state from "../../../common/state";
+import Tools from "../../../common/tools";
+import { removeIgnore, logger, handleError, emojiCollector } from "../common";
+
+export const addBreakRole = async (
+  member: GuildMember,
+  dmChannel: DMChannel
+) => {
+  const confirmationMessage = await dmChannel.send(
+    "It looks like you want a little break! It is understandable you can get it by clicking on the :sloth: emoji below. You can send me !menu again to remove it.\n**BE ADVISED: THIS COMMAND CAN ONLY BE USED EVERY 24 HOURS**"
```
geisterfurz007, line 13:
```suggestion
    "It looks like you want a little break! It is understandable you can get it by clicking on the :sloth: emoji below. You can send me !menu again to remove it.\n**Be advised: You can only toggle your break role every 24 hours**"
```
No need to yell at our poor members ;) Also update the wording to clarify that it's not the menu command that goes on cooldown but toggling the break role.

src/programs/menu/break-handler/addBreak.ts
```diff
@@ -0,0 +1,44 @@
...
+  );
+
+  state.ignoredGroupDMs.push(dmChannel.id);
```
geisterfurz007, line 16: Ignoring the DMs shouldn't be needed here; that was a hastily built thing to ignore incoming messages to YesBot while he was waiting for a naming suggestion (in the past any message would cause the menu to open) and is not relevant for reactions.

src/programs/menu/break-handler/addBreak.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const breakRole = Tools.getRoleByName("Break", member.guild);
+  await confirmationMessage.react("🦥");
+
+  const reaction = await emojiCollector(confirmationMessage, "🦥");
+
+  if (reaction) {
+    try {
+      removeIgnore(dmChannel);
```
geisterfurz007, line 25: See above; this shouldn't be needed

src/programs/menu/break-handler/addBreak.ts
```diff
@@ -0,0 +1,44 @@
...
+      await confirmationMessage.delete();
+      await member.roles.add(breakRole);
+      await prisma.usersOnBreak.create({ data: { userId: member.id } });
+      await dmChannel.send("Enjoy your break!");
+      return;
+    } catch (e) {
+      logger.error("Failed to update break status", e);
+      removeIgnore(dmChannel);
```
geisterfurz007, line 33: See above; this shouldn't be needed

src/programs/menu/break-handler/break-role-events.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+}
+
+@Command({
+  event: DiscordEvent.GUILD_MEMBER_UPDATE,
+  roleNamesRemoved: ["Break"],
+})
+class BreakRemove implements CommandHandler<DiscordEvent.GUILD_MEMBER_UPDATE> {
+  async handle(member: GuildMember): Promise<void> {
+    await prisma.usersOnBreak.delete({ where: { userId: member.id } });
```
geisterfurz007, line 25: I feel like these two could cause problems or at least error messages because entries are created/deleted twice (picking the menu entry creates / deletes the entry and toggles the role causing the same action to be done again). Can we find a way to avoid that duplication (ideally one that doesn't involve just deleting the prisma calls in the menu items)?

src/programs/menu/break-handler/addBreak.ts
```diff
@@ -0,0 +1,44 @@
...
+  );
+
+  state.ignoredGroupDMs.push(dmChannel.id);
+
+  const breakRole = Tools.getRoleByName("Break", member.guild);
+  await confirmationMessage.react("🦥");
+
+  const reaction = await emojiCollector(confirmationMessage, "🦥");
+
+  if (reaction) {
```
geisterfurz007, line 23: Invert this if to reduce nesting and return statements.

src/programs/menu/break-handler/removeBreak.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const breakRole = Tools.getRoleByName("Break", member.guild);
+  const confirmationMessage = await dmChannel.send(
+    "It looks like you're ready to rejoin the server! Once you're ready click on the green check below!"
+  );
+  await confirmationMessage.react("✅");
+
+  const reaction = await emojiCollector(confirmationMessage, "✅");
+
+  if (reaction) {
```
geisterfurz007, line 39: Invert if

src/programs/menu/break-handler/removeBreak.ts
```diff
@@ -0,0 +1,61 @@
...
+  const checkTime = await isCooldownDone(userData);
+
+  if (!checkTime) {
+    await dmChannel.send(
+      "I'm sorry it hasn't been 24 hours since you used this command! If you really want the break role removed you can contact one of our moderators!"
+    );
+    return;
+  }
+
+  state.ignoredGroupDMs.push(dmChannel.id);
```
geisterfurz007, line 29: Not needed, see above.

src/programs/menu/break-handler/removeBreak.ts
```diff
@@ -0,0 +1,61 @@
...
+  const reaction = await emojiCollector(confirmationMessage, "✅");
+
+  if (reaction) {
+    try {
+      await prisma.usersOnBreak.delete({ where: { userId: userData.userId } });
+      await member.roles.remove(breakRole);
+      await dmChannel.send(
+        "Your break role was removed, we're happy to have you back!"
+      );
+      removeIgnore(dmChannel);
```
geisterfurz007, line 46: Not needed.

src/programs/menu/break-handler/removeBreak.ts
```diff
@@ -0,0 +1,61 @@
...
+      return;
+    } catch (e) {
+      logger.error("Failed to update break status", e);
+      removeIgnore(dmChannel);
```
geisterfurz007, line 50: Not needed.

src/programs/menu/break-handler/removeBreak.ts
```diff
@@ -0,0 +1,61 @@
+import { UsersOnBreak } from "@yes-theory-fam/database/client";
```
geisterfurz007, line 1: Filename -> `remove-break.ts`

src/programs/menu/common.ts
```diff
@@ -0,0 +1,65 @@
...
+export const logger = createYesBotLogger("programs", "DmMenu");
+
+export const handleError = async (
+  optionsMessage: Message,
+  dmChannel: DMChannel
+) => {
+  removeIgnore(dmChannel);
+
+  await dmChannel.send(
+    "Because of technical reasons I can only wait 60 seconds for a reaction. I removed the other message to not confuse you. If you need anything from me, just drop me a message!"
```
geisterfurz007, line 25: Having such a specific error message in a function with a very general name `handleError` might be dangerous. Maybe something like `handleReactionTimeout` might be a better fit?

src/programs/menu/common.ts
```diff
@@ -0,0 +1,65 @@
...
+export const isOnBreak = async (userId: string) => {
+  return await prisma.usersOnBreak.findFirst({ where: { userId } });
+};
+
+export const isCooldownDone = async (
+  userData: UsersOnBreak
+): Promise<boolean> => {
+  const twentyFourHours = 24 * 60 * 60 * 1000;
+  const userCoolDownTime = userData.addedAt;
+  return Date.now() - Number(userCoolDownTime) > twentyFourHours;
```
geisterfurz007, line 39:
```suggestion
  return Date.now() - userCoolDownTime.getTime() > twentyFourHours;
```
(Assuming userCoolDownTime is `Date`); does the same but is more readable.

src/programs/menu/common.ts
```diff
@@ -0,0 +1,65 @@
...
+  const reactions = await optionsMessage.awaitReactions({
+    filter,
+    time: 60000,
+    max: 1,
+  });
+  if (reactions.size === 0) throw "No reactions";
+
+  if (!emoji) return reactions.first();
+  return reactions.find((e) => e.emoji.toString() === emoji);
+};
```
geisterfurz007, line 65: This function reads a bit buggy (I didn't test it yet though) 🤔

The most drastic example for how this could go wrong is this: What happens if I provide an emoji that is not in the `allCollectedEmojis` array? The filter doesn't make use of it so it's never collected.

This could also go wrong if I pass emoji as something from the array and the user decides to pick some other option (or adds one themselves) in which case `undefined` will be returned (e.g. emoji === 🦥 and the user adds ✅; the function will try to find an item in the list of ✅ that is a 🦥 which doesn't exist, so undefined).

src/programs/menu/dm-menu.ts
```diff
@@ -0,0 +1,71 @@
...
+      );
+      return;
+    }
+
+    const userOnBreak = await isOnBreak(member.id);
+
+    if (userOnBreak) {
+      await removeBreakRole(member, userOnBreak, dmChannel);
+      return;
+    }
```
geisterfurz007, line 43: Given that we will want to stuff other things into the menu, defaulting to this seems a little drastic (what if a member wants to check their currency while on break for example).

Instead the code should always show all options and the sloth should function as a toggle.

src/programs/menu/nameChange.ts
```diff
@@ -0,0 +1,65 @@
+import { Message, DMChannel } from "discord.js";
```
geisterfurz007, line 1: Filename -> `name-change.ts`

geisterfurz007 (changes requested):
database@0.19.0 is available now as well

src/programs/menu/break-handler/add-break.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (!reaction) {
+    await handleReactionTimeout(confirmationMessage, dmChannel);
+    return;
+  }
+
+  try {
+    await confirmationMessage.delete();
+    await member.roles.add(breakRole);
+    await dmChannel.send("Enjoy your break!");
+    return;
```
geisterfurz007, line 28: `return;` isn't needed here

src/programs/menu/break-handler/add-break.ts
```diff
@@ -0,0 +1,37 @@
...
+  } catch (e) {
+    logger.error("Failed to add break role", e);
+    await textLog(`I could not give <@${member.id}> the break role!`);
+    await dmChannel.send(
+      "Looks like I couldn't give you the break role, I informed the Support team about it, in the meantime you can manually ask one of the Moderators!"
+    );
+    return;
```
geisterfurz007, line 35: `return;` isn't needed here

src/programs/menu/break-handler/remove-break.ts
```diff
@@ -0,0 +1,53 @@
...
+    await handleReactionTimeout(confirmationMessage, dmChannel);
+    return;
+  }
+
+  try {
+    await member.roles.remove(breakRole);
+    await dmChannel.send(
+      "Your break role was removed, we're happy to have you back!"
+    );
+    return;
```
geisterfurz007, line 44: `return;` isn't needed here

src/programs/menu/break-handler/remove-break.ts
```diff
@@ -0,0 +1,53 @@
...
+  } catch (e) {
+    logger.error("Failed to remove break role", e);
+    await textLog(`I could not remove <@${member.id}> break role!`);
+    await dmChannel.send(
+      "Looks like I had a little hiccup trying to remove your role, I've contacted Support about your little issue! Sorry in advance."
+    );
+    return;
```
geisterfurz007, line 51: `return;` isn't needed here

geisterfurz007 (changes requested):

hasanlatch007:
Closing this since it is too old, and will need to be updated, better start from scratch.

PR #592, opened 2021-11-16
Feat: Use handleUserError for event user errors

### LINKED ISSUE #590: Feat: Event error due to location or role should delete the message
Currently `travel-together` is a mess of people spamming `!travel` we could use `Tools.handleUserError` to at least have it clean up after itself keeping the chat a bit cleanish.
If I have the green light for this I can create a PR today :)

hasanlatch007:
Took longer than expected sorry, anyways `refuseWithMessage` was already awaited when called, so I'm assuming you forgot to add `async` to it, should work as intended.

!closes #590


geisterfurz007 (approved):
LGTM!

PR #594, opened 2021-11-17
fix: crash on DM errors

geisterfurz007:


hasanlatch007 (commented):
Shouldn't this also be awaited?
https://github.com/Yes-Theory-Fam/yesbot-ts/blob/0a984baff75b834badc768263354f90fb40e541e/src/programs/filter.ts#L24-L28

geisterfurz007:
> Shouldn't this also be awaited?
> 
> https://github.com/Yes-Theory-Fam/yesbot-ts/blob/0a984baff75b834badc768263354f90fb40e541e/src/programs/filter.ts#L24-L28

Should indeed! Added a couple more async/awaits in the file, thanks!

src/programs/filter.ts
```diff
@@ -43,14 +43,13 @@ class Filter implements CommandHandler<DiscordEvent.MESSAGE> {
 
     if (messageContent.match(/(nigga|nigger)/i)) {
       await message.delete();
-      member
+      await member
         .createDM()
         .then((dm) =>
```
hasanlatch007, line 48:
```suggestion
                .then(async (dm) =>
          await dm.send(
```
This could still crash the bot if the message failed to send if the user has his DM privacy set to friends only

hasanlatch007 (changes requested):

src/programs/filter.ts
hasanlatch007: This was corrected to me on the Discord, thank you Michel!

hasanlatch007 (approved):
LGTM!

PR #595, opened 2021-11-19
fix: discord.js test typescript

geisterfurz007:
discord.js 13.2.0 included a commit that marked all discord.js classes' constructor private. There is an existing issue talking about this although it was closed by its author here: https://github.com/discordjs/discord.js/issues/6798

Since we depend on those constructors for our mocks/test fixtures, I marked all their use with `@ts-expect-error` so we can start updating discord.js again. Doesn't make me happy to do it this way but I don't see any other approach, honestly.

### Notes

1) This branch merges into the renovate update branch! When merging this, please also merge #533 along with it (only when the pipeline goes green of course!).

2) Since this doesn't merge into master, there are no Actions running on this; I ran this locally and it worked but double checking would be good.

### DIFF
--- a/__tests__/event-distribution/events/__snapshots__/member-leave.spec.ts.snap
+++ b/__tests__/event-distribution/events/__snapshots__/member-leave.spec.ts.snap
@@ -20,7 +20,10 @@ Object {
   ],
   "isDirectMessage": false,
   "member": Object {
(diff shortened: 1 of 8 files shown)


adrian-goe (approved):
I don't like this, but I don't see a better way 😕

PR #601, opened 2021-11-23
refactor(*): comment out decorator-test

adrian-goe:


geisterfurz007:
Just to make sure I understand the purpose of this, is it to reduce the log spam?

adrian-goe:
Yes, they trigger each time in production, when the condition meets as well. So, as they are not needed for the production, and is only used as a reference, I'm fine with commenting them out

geisterfurz007 (approved):

PR #637, opened 2022-01-16
Chore: Enabled strict nulls check and fix null errors

### LINKED ISSUE #625: Use strict null checks
Just to note: This is going to be a pest to work through but I genuinely believe it's necessary at the scale YesBot is at now.

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/36dac2af9813c0863f4c789461fe4acf5b2a6a85/tsconfig.json#L14

The tsconfig defines `"strictNullChecks": false` which causes an absurd amount of potential runtime errors to not get picked up by the type checker. To fix this, delete that line from the tsconfig (`strictNullChecks` is enabled through `strict` in the line above) and go on a hunt for null/undefined errors and address them.

hasanlatch007:
First Commit: Worked on errors in `common` folder and enabled the strict nulls check in `tsconfig.ts`

!closes #625

### DIFF
--- a/src/common/moderator.ts
+++ b/src/common/moderator.ts
@@ -8,6 +8,7 @@ import {
 } from "discord.js";
 
 export const isAuthorModerator = (message: Message): boolean => {
+  if (!message.member) return false;
   if (message.member.roles.hoist) {
     return message.member.roles.hoist.name === process.env.MODERATOR_ROLE_NAME;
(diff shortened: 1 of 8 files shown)


geisterfurz007:
Closing as stale; there has been at least some activity in the repo in the last 5 months so I will start from scracth

PR #646, opened 2022-02-05
Upgrade Node in GitHub Workflows

### LINKED ISSUE #644: Upgrade Node in GitHub Workflows
While our Docker container is on Node 16.13, the Actions still run under Node 16.6. This should be updated!

The `.github/workflows` directory contains four files for our actions. This issue concerns only `build.yml`, `test.yml` and `linting.yml` which all define Node 16.6 as their environment. Take this instance for example:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/d85626b90217d8018da5ac3a7565c8ce8e529a9a/.github/workflows/build.yml#L16

This should be updated to Node 16.13 across those three files.

RishiKumarRay:
Fixed #644

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [ 16.x ]
+        node-version: [ 16.13 ]
     steps:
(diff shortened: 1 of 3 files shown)


geisterfurz007 (approved):
LGTM! Thanks a lot :)

PR #667, opened 2022-02-22
removed case sensitive requirement for commands...

joshtrivedi:
... added toLowerCase() to command triggers, and voice on demand

### DIFF
--- a/src/event-distribution/events/message.ts
+++ b/src/event-distribution/events/message.ts
@@ -41,10 +41,10 @@ export const addMessageHandler: AddEventHandlerFunction<
 export const extractMessageInfo: ExtractInfoForEventFunction<
   DiscordEvent.MESSAGE
 > = (message) => {
-  const split = message.content.split(" ");
-  const trigger = split[0].toLowerCase();
-  const subTrigger = split[1];
-
+  const split = message.content.toLowerCase().split(" ");
+  const trigger = split[0] ?? ""
+  const subTrigger = split[1] ?? ""
+  
   return withMessageRelatedInfo(message, message.member, (channelId) => [
     channelId,
     trigger,
--- a/src/event-distribution/events/message.ts
+++ b/src/event-distribution/events/message.ts
@@ -42,7 +42,7 @@ export const extractMessageInfo: ExtractInfoForEventFunction<
   DiscordEvent.MESSAGE
 > = (message) => {
   const split = message.content.split(" ");
-  const trigger = split[0];
+  const trigger = split[0].toLowerCase();
   const subTrigger = split[1];
 
   return withMessageRelatedInfo(message, message.member, (channelId) => [
--- a/src/event-distribution/events/message.ts
+++ b/src/event-distribution/events/message.ts
@@ -42,9 +42,9 @@ export const extractMessageInfo: ExtractInfoForEventFunction<
   DiscordEvent.MESSAGE
 > = (message) => {
   const split = message.content.toLowerCase().split(" ");
-  const trigger = split[0] ?? ""
-  const subTrigger = split[1] ?? ""
-  
+  const trigger = split[0] ?? "";
+  const subTrigger = split[1] ?? "";
+
   return withMessageRelatedInfo(message, message.member, (channelId) => [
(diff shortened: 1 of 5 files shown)


src/event-distribution/events/message.ts
```diff
@@ -42,7 +42,7 @@ export const extractMessageInfo: ExtractInfoForEventFunction<
   DiscordEvent.MESSAGE
 > = (message) => {
   const split = message.content.split(" ");
```
hasanlatch007, line 44:
```suggestion
  const split = message.content.toLowerCase().split(" ");
  const trigger = split[0];
  const subTrigger = split[1];
```
The way you've done is basically having it 50/50, any command send the trigger is lowerCased but not the subtrigger, for example `!group Join` would not work.

hasanlatch007 (changes requested):
This actually breaks some of our commands, for example `!timeoutTimer`, it is a trigger with one capitalized letter, with the `toLowerCase()` in [src/event-distribution/events/message.ts](https://github.com/Yes-Theory-Fam/yesbot-ts/pull/667/files#diff-77b198c2b3d38af921c099137af5327d91540a4b70eac40f4391ab96d0de43d8) every message is extracted in lower case and won't be picked up by the new event handler system.

joshtrivedi (commented):
reviewed, please verify

src/event-distribution/events/message.ts
```diff
@@ -41,9 +41,9 @@ export const addMessageHandler: AddEventHandlerFunction<
 export const extractMessageInfo: ExtractInfoForEventFunction<
   DiscordEvent.MESSAGE
 > = (message) => {
-  const split = message.content.split(" ");
-  const trigger = split[0];
-  const subTrigger = split[1];
+  const split = message.content.toLowerCase().split(" ");
+  const trigger = split[0] ?? "";
+  const subTrigger = split[1] ?? "";
```
hasanlatch007, line 46: ` ?? ""` Is useless here.

hasanlatch007 (changes requested):
This still breaks commands that have a capital letter like `!timeoutTimer` I would recommend to add this in [src/event-distribution/events/message.ts](https://github.com/Yes-Theory-Fam/yesbot-ts/pull/667/files#diff-77b198c2b3d38af921c099137af5327d91540a4b70eac40f4391ab96d0de43d8) at line `32` and `33`

```suggestion
  const trigger = options.trigger?.toLowerCase() ?? "";
  const subTrigger = options.subTrigger?.toLowerCase() ?? "";
```

src/event-distribution/events/message.ts
joshtrivedi: that was in case there are no sub triggers, it was still passing null

joshtrivedi (commented):
check again shawty

joshtrivedi:
something fails help

geisterfurz007:
> something fails help

The testsuite creates snapshot files (basically taking the output, throwing it in a file and saving that) for various outputs of the event distribution. On subsequent testruns, the actual outputs are compared against the file and if the content doesn't match the new result, the test fails.  

Since the whole thing is now case insensitive, the output has changed (the triggers extracted from a message are now converted to lowercase, altering the HandlerInfo returned for some messages in the tests). You can update the tests using `yarn test -u` telling jest that the changes were intentional causing those snapshot files to be written again. Commit those and push them, then the pipeline should go green again 👍

geisterfurz007 (changes requested):
During testing one interaction turned out to still be case-sensitive; the group ping isn't handled by a trigger but a general check if the message.content contains `@group` which currently is not case-insensitive.

A `.toLowerCase()` near the top of group-ping.ts should fix this.

After that, LGTM!

src/event-distribution/events/message.ts
hasanlatch007: I don't know if I would need it for when strictNullsCheck is enabled so Ill keep it.

geisterfurz007 (approved):
LGTM!

hasanlatch007 (approved):
Also LGTM!

PR #692, opened 2022-04-10
Birthdays in YesBot

geisterfurz007:
This PR aims to replace BirthdayBot with YesBot.

Relevant changes made:

 - New !setTimezone command for members where timezones were never recorded
 - New Cron job running every 15 minutes to add / remove birthday roles
 - Currently only reports intended changes to the logging channel (to be reverted before merging)


geisterfurz007:
Stale and overall not fleshed out at all...

Issue #551, opened 2021-10-20
dailychallenge ping doesn't ping group


geisterfurz007:
The dailychallenge ping as well as the yestheoryuploads ping from the bot don't appear to actually cause a ping since sunday requiring a manual one instead:

![grafik](https://user-images.githubusercontent.com/26303198/138053794-2135d8ab-7f6a-4e84-a8e3-a17b04691de6.png)

There are no visible errors in the log. This needs investigation and fix.

joshtrivedi:
I think it has to do something with GroupPingSetting, since when a member pings the group it tags, but not when yes bot does
![image](https://user-images.githubusercontent.com/50212531/155967545-c9fd772a-8eb9-48a5-95d3-f0cf8e7de049.png)

geisterfurz007:
In that case, the behaviour should be consistent which it isn't; take today's daily challenge for example where everything worked as expected: 

![grafik](https://user-images.githubusercontent.com/26303198/155975547-883c4e2c-cc44-4b2d-bd9a-23a122e2f92e.png)

I'm gonna release an update to YesBot later which will include #664; @hasanlatch007 had done some investigation in the past and it seemed that some events were just being dropped by discord.js so sending everything in one message could be an option.  
If that doesn't turn out to be a solution, we can ramp up the logging, possibly even within discord.js to see if anything messes up in there but fingers crossed that's not needed!

hasanlatch007:
> I think it has to do something with GroupPingSetting, since when a member pings the group it tags, but not when yes bot does ![image](https://user-images.githubusercontent.com/50212531/155967545-c9fd772a-8eb9-48a5-95d3-f0cf8e7de049.png)

What Michel said, also if you try to ping a group that has the `GroupPingSetting` set it would result in an error, saying `This group is only pingable by {SETTING}`, for now, we're testing sending everything in one message.

joshtrivedi:
> > I think it has to do something with GroupPingSetting, since when a member pings the group it tags, but not when yes bot does ![image](https://user-images.githubusercontent.com/50212531/155967545-c9fd772a-8eb9-48a5-95d3-f0cf8e7de049.png)
> 
> What Michel said, also if you try to ping a group that has the `GroupPingSetting` set it would result in an error, saying `This group is only pingable by {SETTING}`, for now, we're testing sending everything in one message.

So by that you mean including the group tag in the embed or sending the whole thing as a message.

geisterfurz007:
[...]

PR #707, opened 2022-05-04
added 1s delay to auto ping

hasanlatch007:
Possible fix for #551


geisterfurz007 (approved):
Yup, let's give this a shot

PR #708, opened 2022-05-07
Fix: Force ping needed groups

hasanlatch007:
Probable fix for #551

### DIFF
--- a/src/common/tools.ts
+++ b/src/common/tools.ts
@@ -12,11 +12,13 @@ import {
   Snowflake,
   TextChannel,
   User,
+  Util,
 } from "discord.js";
 import { textLog } from "./moderator";
 import { createYesBotLogger } from "../log";
 import prisma from "../prisma";
 import axios from "axios";
+import { findManyRequestedGroups } from "../programs/group-manager/common";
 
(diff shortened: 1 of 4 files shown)


geisterfurz007 (changes requested):
This PR introduces several code duplications; there should be one function that takes in a group name and TextChannel and sends the group ping. As it stands now, code would have to be adapted thrice if for example the group ping message was to be adapted.

hasanlatch007:
> This PR introduces several code duplications; there should be one function that takes in a group name and TextChannel and sends the group ping. As it stands now, code would have to be adapted thrice if for example the group ping message was to be adapted.

Damn i was hoping we'd ignore code duplication for this bug :') will rework it

geisterfurz007 (changes requested):
The end of the group-ping logic can be replaced by this method as well:

https://github.com/Yes-Theory-Fam/yesbot-ts/blob/70690969451b9fc379efd41ab304223ee586cb07/src/programs/group-manager/user-commands/group-ping.ts#L116-L130

geisterfurz007 (approved):

PR #727, opened 2022-06-06
Add some missing awaits and catch calls

geisterfurz007:
There are still some instances where YesBot restarts on errors deleting messages so this PR adds some awaits and catch clauses in places where messages are deleted so it's not a problem if a user deletes their message faster than YesBot can.

### DIFF
--- a/src/common/tools.ts
+++ b/src/common/tools.ts
@@ -88,8 +88,8 @@ class Tools {
 
   static async handleUserError(message: Message, reply: string) {
     message.reply(reply).then((msg) => {
-      message.delete();
-      setTimeout(() => msg.delete(), 10000);
+      message.delete().catch(() => {});
+      setTimeout(() => msg.delete().catch(() => {}), 10000);
     });
   }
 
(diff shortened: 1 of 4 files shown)


hasanlatch007:
Missing linting

hasanlatch007 (approved):
Couldn't spot any more missing catch and awaits. LGTM!

hasanlatch007:
Requires rebase

geisterfurz007:
> Couldn't spot any more missing catch and awaits. LGTM!

Oh WebStorm spotted plenty but the most recent crashes seemed to have been caused by errors in `Message#delete` calls, so those were my priority here. We can clean up the rest slowly but surely. The problematic ones will be easy to spot soon once null safety isn't an issue anymore!

> Requires rebase

Done; merging then :)

hasanlatch007:
> > Couldn't spot any more missing catch and awaits. LGTM!
> 
> Oh WebStorm spotted plenty but the most recent crashes seemed to have been caused by errors in `Message#delete` calls, so those were my priority here. We can clean up the rest slowly but surely. The problematic ones will be easy to spot soon once null safety isn't an issue anymore!
> 
> > Requires rebase
> 
> Done; merging then :)

Should probably start using WebStorm then, I was going through major files 👀

PR #738, opened 2022-06-17
Kubernetes Deployment

### LINKED ISSUE #723: Kubernetes Deployments
The current docker-compose setup does its job but moving the bot to Kubernetes might have some benefits in the future like updates without downtime.

This is certainly not a pressing issue however.

geisterfurz007:
Closes #723


geisterfurz007:
The K8s Deployment of YesBot will be postponed for a while. After a discussion with @sklirg we found that the main driver for me to do this (i.e. better access to a joined database between YesBot and other projects) could and probably should be done in a different manner. That is not to say it will never happen; just not in the near future.

PR #761, opened 2022-07-03
feat: limit number of countries per travel ticket to 3

### LINKED ISSUE #760: Limit number of pinged countries in travel tickets to 3

geisterfurz007:
Closes #760


hasanlatch007 (approved):
LGTM! Just noting, if the country selected is the USA, it takes the bot a little time before it registers the user has reacted with a green checkmark, I'll let the merging to you if you don't want to optimize it more and leave it be. Other than that all is good!

PR #825, opened 2022-10-08
feat: More slash commands! (`/deadchat`, `/profile`, `/resources`, `/video`)

sm-sami:
This PR converts `!deadchat`, `!profile`, `!resources` and `!video` to their respective slash commands.

### DIFF
--- a/src/programs/deadchat.ts
+++ b/src/programs/deadchat.ts
@@ -9,9 +9,9 @@ const isDead = async (
   interaction: ChatInputCommandInteraction
 ): Promise<boolean> => {
   const channel = interaction.channel as TextChannel;
-  const lastMessages = (await channel.messages.fetch({ limit: 2 })).values();
-  const lastMessage = [...lastMessages][1];
+  const lastMessage = (await channel.messages.fetch({ limit: 1 })).first();
 
+  if (!lastMessage) return true;
   return Date.now() - lastMessage.createdTimestamp > thirtyMinutes;
 };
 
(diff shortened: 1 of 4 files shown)


src/programs/deadchat.ts
```diff
@@ -1,33 +1,33 @@
...
-const isDead = async (message: Message): Promise<boolean> => {
-  const lastMessages = (
-    await message.channel.messages.fetch({ limit: 2 })
-  ).values();
+const isDead = async (
+  interaction: ChatInputCommandInteraction
+): Promise<boolean> => {
+  const channel = interaction.channel as TextChannel;
+  const lastMessages = (await channel.messages.fetch({ limit: 2 })).values();
   const lastMessage = [...lastMessages][1];
```
geisterfurz007, line 13: This is a sneaky one; previously the bot had to skip the message containing the command when determining whether chat was dead or not. Since the slash command does not count as a message, this is not needed anymore so the code needs to be changed to actually look for the last message (instead of the second to last message as it was previously):

```suggestion
  const lastMessage = (await channel.messages.fetch({ limit: 1 })).first();
  if (!lastMessage) return true;

```

geisterfurz007 (changes requested):
The rest looks perfect though!

src/programs/deadchat.ts
sm-sami: Ah, my bad for the oversight.

geisterfurz007: No worries, that's why we do reviews :)

PR #828, opened 2022-10-25
feat: automatically add role "from strangers to friends"

L-390:


src/programs/strangers-to-friends.ts
```diff
@@ -0,0 +1,23 @@
+import { Command, CommandHandler, DiscordEvent } from "../event-distribution";
+import { GuildMember, Message } from "discord.js";
+
+@Command({
+  event: DiscordEvent.MESSAGE,
+  channelNames: ["from-strangers-to-friends"],
```
geisterfurz007, line 6: Prefer creating a new entry in `src/collections/chat-names.ts` over setting it as string here; that makes it easier for the future in case there might be new features in that channel (wink, wink, nudge, nudge).

src/programs/strangers-to-friends.ts
```diff
@@ -0,0 +1,23 @@
...
+  description:
+    "This handler is for autoassigning the from strangers to friends roles to everyone mentioned in a post.",
+})
+class FromStrangersToFriends implements CommandHandler<DiscordEvent.MESSAGE> {
+  async handle(message: Message): Promise<void> {
+    let users = message.mentions?.members;
```
geisterfurz007, line 12: Prefer `const` for values that don't change

src/programs/strangers-to-friends.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    users?.forEach((user: GuildMember) => {
```
geisterfurz007, line 14: Prefer a for-of loop over array foreach (arguably, this is very subjective but for-of is used in pretty much the entirety of YesBot's codebase, so I'd rather see it used here as well); you can use something like `if (!users) return;` on a line above that then to avoid iterating over undefined / null.

src/programs/strangers-to-friends.ts
```diff
@@ -0,0 +1,23 @@
...
+      if (user.roles?.cache.has("499143174271270913")) {
```
geisterfurz007, line 15: Don't hardcode the roleid; that makes code impossible to test on testservers. Instead, look up the role by name (there is `Tools.getRoleByName` in `src/common/tools.ts`) when the handler is run.

src/programs/strangers-to-friends.ts
```diff
@@ -0,0 +1,23 @@
...
+        return;
+      } else {
```
geisterfurz007, line 17: An empty if block is better expressed by inverting the condition and moving the else logic there:

```ts
if (something) {}
else { console.log('Some code here'); }
```
is equal to
```ts
if (!something) { console.log(`Some code here`); }
```

src/programs/strangers-to-friends.ts
```diff
@@ -0,0 +1,23 @@
...
+        user.roles.add("499143174271270913");
+        return;
```
geisterfurz007, line 19: `return;` as last statement in a function is unnecessary.

geisterfurz007 (changes requested):
This looks promising! I have a couple points on code-style and a couple minor nitpicks however :)

L-390:
Hope I didn't forget any of your suggestsion :)

src/programs/strangers-to-friends.ts
```diff
@@ -0,0 +1,28 @@
...
+  channelNames: [ChatNames.STRANGERS_TO_FRIENDS],
+  description:
+    "This handler is for autoassigning the from strangers to friends roles to everyone mentioned in a post.",
+})
+class FromStrangersToFriends implements CommandHandler<DiscordEvent.MESSAGE> {
+  async handle(message: Message): Promise<void> {
+    if (!message.guild) return;
+    const users = message.mentions.members;
+    const role = Tools.getRoleByName(
+      "From Strangers To Friends :camera:",
```
geisterfurz007, line 17: The emoji must be a literal emoji

```suggestion
      "From Strangers To Friends 📸",
```

src/programs/strangers-to-friends.ts
```diff
@@ -0,0 +1,28 @@
...
+      message.guild
+    );
+
+    if (!users || !role) return;
+    for (const user of users.values()) {
+      if (!user.roles?.cache.has(role.id)) {
+        user.roles.add(role.id);
```
geisterfurz007, line 24: `await` asynchronous function calls so errors are caught in the event handler and don't bubble up to crash the bot.

```suggestion
        await user.roles.add(role.id);
```

geisterfurz007 (changes requested):
Two more little nitpicks but then we should be good!

geisterfurz007 (approved):
LGTM! Thanks a bunch!

PR #829, opened 2022-10-28
feat: @someone command as slash command

L-390:


src/programs/someone-tag.ts
```diff
@@ -8,90 +16,64 @@ const QUESTION_SHEET_ID: string =
...
-  channelNames: ["chat", "chat-too", "4th-chat", "chat-v"],
-  description: "This handler is for the someone command.",
+  event: DiscordEvent.SLASH_COMMAND,
+  root: "someone",
+  description: "Ping a random memeber.",
+  options: [
+    {
+      type: ApplicationCommandOptionType.Boolean,
+      name: "Online",
+      description: "Should the person be Online.",
```
geisterfurz007, line 26: I am not sure about this description; to me this reads like "If false, no online member will be pinged" which is not accurate. Maybe something like "If true, ensures the pinged member is online" works better? Not sure but the phrasing should be improved, I feel like.

src/programs/someone-tag.ts
```diff
@@ -8,90 +16,64 @@ const QUESTION_SHEET_ID: string =
...
+class SomeoneTag implements CommandHandler<DiscordEvent.SLASH_COMMAND> {
+  async handle(interaction: ChatInputCommandInteraction): Promise<void> {
+    const allow = await isAllowed(interaction.user);
 
     if (!allow) {
-      await Tools.handleUserError(
-        message,
-        "You have already used this command today!"
-      );
+      interaction.reply("You have already used this command today!");
```
geisterfurz007, line 36: - `await` async functions
- Error responses should be ephermal

```suggestion
      await interaction.reply({ content: "You have already used this command today!", ephermal: true });
```

src/programs/someone-tag.ts
```diff
@@ -8,90 +16,64 @@ const QUESTION_SHEET_ID: string =
...
-  allowedRoles: ["Seek Discomfort"],
-  channelNames: ["chat", "chat-too", "4th-chat", "chat-v"],
-  description: "This handler is for the someone command.",
+  event: DiscordEvent.SLASH_COMMAND,
+  root: "someone",
+  description: "Ping a random memeber.",
+  options: [
+    {
+      type: ApplicationCommandOptionType.Boolean,
+      name: "Online",
```
geisterfurz007, line 25: The name must be lowercase (otherwise it results in a runtime error when starting the bot):

```suggestion
      name: "online",
```

src/programs/someone-tag.ts
```diff
@@ -8,90 +16,64 @@ const QUESTION_SHEET_ID: string =
...
-          message.channel as TextChannel
-        );
-      }
-    }
+    const { member } = interaction;
+    if (!member) return;
 
-    await message.delete();
+    const target = await getTarget(
+      interaction.options.getBoolean("Online"),
```
geisterfurz007, line 44: Consequence from the change in the options above.

```suggestion
      interaction.options.getBoolean("online"),
```

src/programs/someone-tag.ts
```diff
@@ -8,90 +16,64 @@ const QUESTION_SHEET_ID: string =
...
+      interaction
+    );
+    const question = await getQuestion();
+    if (!target) {
+      await interaction.reply(
+        "There were no available users to ping! This is embarrassing. How could this have happened? There's so many people on here that statistically this message should never even show up. Oh well. Congratulations, I guess."
+      );
```
geisterfurz007, line 51:
```suggestion
      await interaction.reply({
        content: "There were no available users to ping! This is embarrassing. How could this have happened? There's so many people on here that statistically this message should never even show up. Oh well. Congratulations, I guess.",
        ephermal: true,
      });
```

src/programs/someone-tag.ts
```diff
@@ -8,90 +16,64 @@ const QUESTION_SHEET_ID: string =
...
+    } else {
+      await sendMessage(
+        member.user as User,
+        target,
+        question,
+        interaction.channel as TextChannel
+      );
+    }
```
geisterfurz007, line 59: All interactions have to be replied to (otherwise there is an error message); so this will need something like `interaction.reply({ content: "Done!", ephermal: true });` added to it.

Some future discord.js version will also allow deleting the ephermal messages buuuut I haven't updated to that yet so this will have to do for now!

src/programs/someone-tag.ts
L-390: not sure if VSC is right, but it gives me an error saying that "ephermal" doesn't exist on type "InteractionReplyOptions" .

src/programs/someone-tag.ts
L-390: Same thing here, "ephermal" doesn't exist on type 'InteractionReplyOptions | MessagePayload'.

src/programs/someone-tag.ts
L-390: also same thing with ephermal not existing

L-390:
nvm all the comments I wrote :), it's "ephemeral" instead of "ephermal"

geisterfurz007:
Oh whoops, sorry about that!

geisterfurz007 (approved):
LGTM, thanks a lot!

PR #841, opened 2022-11-12
Replaced user mention with user id when generating two slash commands

julisod:
Result:
image


geisterfurz007 (changes requested):
Yeah, this looks good! Could you run `yarn run lint:fix` and commit the results from that, so that everything is nice and tidy?

geisterfurz007 (approved):
Perfect, thanks! That's gonna make things a lot smoother for us :)

PR #842, opened 2022-11-13
fixed up and down commands

### LINKED ISSUE #728: Voice Up doesn't up more than voiceLimit + 1
I'm not the sharpest tool in the shed for not thinking about this
image

hasanlatch007:
!closes #728

### DIFF
--- a/src/programs/voice-on-demand.ts
+++ b/src/programs/voice-on-demand.ts
@@ -135,10 +135,10 @@ const getShrinkLimit = (channel: VoiceChannel) =>
   Math.max(2, channel.members.size);
 
 const getUpLimit = (channel: VoiceChannel) =>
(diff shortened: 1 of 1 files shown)


geisterfurz007 (changes requested):
This reverts to a [bug that previously existed](https://github.com/Yes-Theory-Fam/yesbot-ts/pull/713):

If you had 7 people in a room and limited it to 2, you had to up quite a bit.

I believe something like `Math.min(maxLimit, Math.max(channel.members.size, channel.userLimit))` should cover the upLimit just nicely.

The downLimit should be fine like you changed it her.

geisterfurz007:
Closing this in favour of the upcoming slash-command rework which will include this fix

PR #843, opened 2022-11-13
Fix: race condition

### LINKED ISSUE #736: Race-Condition when starting breaking role reactions
Sometimes when the bot starts, the guilds might not be fully loaded yet causing failure to resolve the color reaction role.

It would need a way to ensure the guilds have been loaded before the ready handlers are running to make sure everything is *really* loaded and not just have the client logged in.

hasanlatch007:
This could be a possible fix for #736 since we have two `ready.ts` that handle the bot ready event, we can make the old event take resolving the guild before heading to the new event distribution system. I'm assuming that when you said color selection, you meant the nitro colors since they use the new `ready.ts` loading the old one first doesn't seem like a lousy idea 🤔 

Hopefully !closes #736

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -118,7 +118,7 @@ bot.on(
 );
 bot.on("ready", async () => {
   await ready(bot).catch((error) =>
-      logger.error("Error in legacy ready handler: ", error)
+    logger.error("Error in legacy ready handler: ", error)
   );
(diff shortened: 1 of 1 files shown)


geisterfurz007 (approved):
Yeah, let's see if this helps! LGTM!

PR #844, opened 2022-11-13
Feat: Thread for Highlighted Quotes

### LINKED ISSUE #700: Move "Highlighted" Quotes from pins to thread
Pinning in quotes is limited to 50 messages and hard to monitor or control manually. A fix would be to move the messages with the correct voting ratio to be reposted within a thread of #polls that people can decide to join for those notifications.

hasanlatch007:
!closes #700
Would love some feedback, feel like it could be improved.


hasanlatch007:
Too old would rather rework it from scratch

Issue #720, opened 2022-06-05
Approve shoutout tickets


geisterfurz007:
Currently the process to approve a shoutout process looks like this:

1. Copy the user's name / tag
2. Go to #shoutout settings
3. Create new permissions for that user
4. Send message in the ticket channel letting them know that they can post now
5. Wait for them to post
6. Remove permissions from #shoutout
7. Close the ticket

This could possibly be automated with some sort of approval command that handles at least steps 1 through 4. In a later stage this might be extended to also cover 5-7 but I am not sure yet.

christolis:
Has anybody worked on this? Also, I recommend setting a default 24-hour time space before the user has their permission automatically revoked for safety reasons.

PR #966, opened 2023-03-24
Approve shoutout tickets

christolis:
Resolves issue #720.

### Things to do
- [x] Automatically remove permissions once the privileged user posts a message in the shoutouts channel.
- [x] Implement a system so that privileged users lose the permission to post a message after a day.
- [x] Improve the messages sent by the bot so that they fit more with what the Yes Theory Bot would send (will ~~probably~~ need external input on this).
- [x] Think about an efficient way to detect the user who opened the ticket from the ticket text channel instead of manually inputting it as an argument.
- [x] ~~Make it so that the command is only executable by users with a role that authorizes them to do so (probably the Support role but I am not entirely sure)~~ covered already!


christolis:
Hey @geisterfurz007, which role(s) should allow users to execute the command? 🤔

geisterfurz007:
@christolis the permissions for slash commands can be configured directly in Discord (Server Settings > Integrations > Your Bot), so you don't have to worry about that :)

christolis:
Good day @geisterfurz007! As for this task:
> _(maybe)_ Think about an efficient way to detect the user who opened the ticket from the ticket text channel instead of manually inputting it as an argument.

Do you think we should rename individual shoutout tickets to a format like `#shoutout-{userId}` instead of `#shoutout-{usernametag}` in order to reduce the /shoutout command to `/shoutout toggle` and automatically get the target user depending on where the command was executed? Or should we leave it as it is?

Also, whenever you have time (or if anybody else can help me with it), I'll need some help implementing this:
> Improve the messages sent by the bot so that they fit more with what the Yes Theory Bot would send (will probably need external input on this).

geisterfurz007:
> Do you think we should rename individual shoutout tickets to a format like `#shoutout-{userId}` instead of `#shoutout-{usernametag}`

Yes, absolutely! That also fixes a (fairly uncommon but still possible) bug where someone can change their username while having a ticket open and having more than one ticket open at a time and it removes the need for sanitizing the username/tag.

Regarding the messages: Is this only regarding phrasing? If yes, feel free to put any text that you feel is reasonably polite and to the point; I will go through the texts during review again either way. If not, I am not sure I understand what you mean.

src/programs/shoutout-perms.ts
```diff
@@ -0,0 +1,177 @@
...
+      await interaction.reply({
+        content: `Couldn't find the <#${ChatNames.SHOUTOUTS}> channel!`,
+        ephemeral: true,
+      });
+      return;
+    }
+
+    // Check whether there's an instance.
+    // If there is an existing perm overwrite, remove it
+    const sendPerms = getSendPerms(shoutouts, user);
```
geisterfurz007, line 98: Earlier I mentioned removing the getSendPerms in favour of a shorter deletePerms function; since this is the only place where we actually want to know if that user has those permissions, we can check it with this:
```ts
const perms = await shoutouts.permissionsFor(user);
const hasSendPerms = perms?.has(PermissionsBitField.Flags.SendMessages) ?? false;
```

src/programs/shoutout-perms.ts
```diff
@@ -0,0 +1,177 @@
...
+    const user = bot.users.resolve(data.userId);
+
+    if (user === null || channel === null) return;
+    const sendPerms = getSendPerms(channel, user);
+
+    if (sendPerms) await deletePerms(user.id, sendPerms);
+  }
+}
+
+const logger = createYesBotLogger("shoutouts", ShoutoutTimer.name);
```
geisterfurz007, line 177: The logger is unused and can be removed

src/programs/shoutout-perms.ts
```diff
@@ -0,0 +1,177 @@
...
+        ephemeral: true,
+      });
+      return;
+    }
+
+    // Check whether there's an instance.
+    // If there is an existing perm overwrite, remove it
+    const sendPerms = getSendPerms(shoutouts, user);
+    if (sendPerms) {
+      deletePerms(user.id, sendPerms);
```
geisterfurz007, line 100: returned `Promise`s *must* be awaited; since Node 16 any errors that comes up in a Promise that doesn't have either a `.catch` or an `await` inside a `try/catch` will terminate the process. All handlers are wrapped in `try/catch` somewhere deep in the "core" of YesBot but for that to work, `await` is needed.
```suggestion
      await deletePerms(user.id, sendPerms);
```

geisterfurz007 (changes requested):
For a first run, this already looks really nice, good job! A couple comments though; if you have any questions on them, let me know!

src/programs/shoutout-perms.ts
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    // Check whether there's an instance.
+    // If there is an existing perm overwrite, remove it
+    const perms = await shoutouts.permissionsFor(user);
+    const hasSendPerms =
+      perms?.has(PermissionsBitField.Flags.SendMessages, false) ?? false;
+
+    if (hasSendPerms) {
+      await channel.permissionOverwrites.delete(user);
```
geisterfurz007, line 84:
```suggestion
      await shoutouts.permissionOverwrites.delete(user);
```

geisterfurz007 (changes requested):
Cheeky copy-paste bug hid itself :)

geisterfurz007 (approved):
LGTM! Thanks a lot for the contribution! The team will be happy to not need to click through 5 menus again when dealing with shoutouts :)

christolis:
I'm glad that that's the case! I'd love to work on more issues and things that you guys might want added to the project! 😇

PR #974, opened 2023-04-20
[WIP] Automate Nitro Booster monthly cleanup

christolis:
This pull request aims to automate the tedious monthly manual task of removing the Nitro booster color roles from all members of the server, by removing the reactions from the reaction message where members get their color roles from and by announcing the monthly change to the Nitro boosters!

# TODO
- [x] Make the messages a bit nicer, if necessary!


src/programs/nitro-colors.ts
```diff
@@ -139,6 +140,60 @@ class NitroColorSelector implements CommandHandler<DiscordEvent.REACTION_ADD> {
   }
 }
 
+export class RoleResetCron {
+  static init() {
+    const MONTHLY_CRON = "0 0 1 * *";
+    const TEST_CRON = "*/1 * * * *"; // Every 1 minute, for testing purposes
```
christolis, line 146: @geisterfurz007 I have a feeling that this could be improved by automatically picking `TEST_CRON` depending on whether the bot is running in a test server or not (as far as I remember, there are some environment variables that could possibly help with this). What do you think?

geisterfurz007: Sorry, this took ages for me to get to! Yes, you can rely on `process.env.NODE_ENV` to have the value `"production"` when built for production. By default it has the value `"development"` and when the tests are running it has the value `"test"`. So easiest to test on `process.env.NODE_ENV === "production"` to determine which schedule to use!

src/programs/nitro-colors.ts
```diff
@@ -139,6 +140,60 @@ class NitroColorSelector implements CommandHandler<DiscordEvent.REACTION_ADD> {
...
+      const channel = bot.guilds
+        .resolve(process.env.GUILD_ID)
+        ?.channels.cache.find(
+          (c) => c.name === "pick-your-color"
+        ) as TextChannel;
+
+      // Remove all messages sent by the bot
+      channel.messages
+        .fetch({ limit: 5 })
+        .then((messages) => {
```
geisterfurz007, line 167: There is a lot of nesting going on here; instead of using `.then()`, use `await` to avoid deeply nested callbacks.

src/programs/nitro-colors.ts
```diff
@@ -139,6 +140,60 @@ class NitroColorSelector implements CommandHandler<DiscordEvent.REACTION_ADD> {
...
+    const MONTHLY_CRON = "0 0 1 * *";
+    const TEST_CRON = "*/1 * * * *"; // Every 1 minute, for testing purposes
+
+    // Schedule a cron task every month
+    cron.schedule(MONTHLY_CRON, async () => {
+      // Remove color roles from each user
+      nitroRolesCache.forEach((role) => {
+        role.members.forEach(async (member) => {
+          await member.roles.remove(role);
+        });
```
geisterfurz007, line 154: Avoid placing `async` code inside of `forEach`; the promises inside are "dangling" and rejections inside of them are uncaught. Instead either use a `for-of` loop, or `.map` an array of values to an array of promises and then use `await Promise.all()` to ensure all errors are handled inside the correct callstack.

src/programs/nitro-colors.ts
```diff
@@ -139,6 +140,60 @@ class NitroColorSelector implements CommandHandler<DiscordEvent.REACTION_ADD> {
...
+              });
+            }
+          });
+        })
+        .then(async () => {
+          // TODO: Make this message a bit nicer!
+          // Let Nitro boosters know about the new month's change!
+          await channel.send({
+            content:
+              "@Nitro Booster It is time to pick a new color for the new month!",
```
geisterfurz007, line 187: I am not sure if this ping here would work (since the API has no chance of knowing where the role name ends). You can use `Tools.getRoleByName("Nitro Booster", guild)` to get the actual role and use that in the string `${boosterRole} It is time [...]` which should yield a proper ping.

src/programs/nitro-colors.ts
```diff
@@ -139,6 +140,60 @@ class NitroColorSelector implements CommandHandler<DiscordEvent.REACTION_ADD> {
   }
 }
 
+export class RoleResetCron {
```
geisterfurz007, line 143: To avoid this file getting *huge* (my usual limit is around 150 lines), it's cleaner to create a new folder called `nitro-colors`, move this new reset logic in a separate file into that folder and rename *this* file to something sensible and move it in there as well.

geisterfurz007 (changes requested):

geisterfurz007:
I also think it makes sense to also include some logic for editing the roles between seasons to avoid pinging people when that change hasn't been made yet by a moderator. Let me know if that's something you'd like to do in this PR as well; if it is, I can pass you the list of color codes and names here (it won't take two months this time, I hope!).

src/programs/nitro-colors.ts
christolis: Done!

src/programs/nitro-colors.ts
christolis: Should be fixed with the newest commits! 🙌

src/programs/nitro-colors.ts
christolis: Don't worry about it! I did end up using `process.env.NODE_ENV` to control which cron expression to use, and it appears to be working perfectly!

src/programs/nitro-colors.ts
christolis: Thank you for the suggestion! I will keep this in mind next time I need to deal with asynchronous code inside `forEach` loops. It should be changed now!

src/programs/nitro-colors.ts
christolis: Done! I renamed the original file to `index.ts`.

christolis:
> I also think it makes sense to also include some logic for editing the roles between seasons to avoid pinging people when that change hasn't been made yet by a moderator. Let me know if that's something you'd like to do in this PR as well; if it is, I can pass you the list of color codes and names here (it won't take two months this time, I hope!).

Sure thing! Although, along with the list of the color codes and names, could you please provide a more detailed explanation of how this will work?

geisterfurz007:
> Sure thing! Although, along with the list of the color codes and names, could you please provide a more detailed explanation of how this will work?

Absolutely! For each (northern hemisphere) season, we have 5 shades of the same colors red, green, yellow, blue and orange. Winter colours are available from December to February, Sprint from March to May, Summer from June to August and Autumn from September till November (all ranges inclusive).

So essentially, at the start of the month, remove everyone's roles, check if it's time for new colours, if yes, update the names and colors of the five roles, if not, leave them untouched. Then once that is finished, post the message in the channel that new stuff is available.

Colours in JSON because that's possibly easiest to make use of:

```json
{
  "summer": {
    "red": { "name": "Icecream", "color": "#e0234a" },
    "green": { "name": "Meadow", "color": "#08bd56" },
    "yellow": { "name": "Sunshine", "color": "#e0d41a" },
    "blue": { "name": "Cloudless", "color": "#0fbdb5" },
    "orange": { "name": "Beachball", "color": "#ff491f" },
  },
  "autumn": {
    "red": { "name": "Bonfire", "color": "#cf453a" },
    "green": { "name": "Overgrown", "color": "#7ba702" },
    "yellow": { "name": "Pumpkin", "color": "#dfac3a" },
    "blue": { "name": "Moonshine", "color": "#1a8391" },
    "orange": { "name": "Foliage", "color": "#d1652b" },
  },
  "winter": {
    "red": { "name": "Candlelight", "color": "#b10f0a" },
    "green": { "name": "Pinetree", "color": "#1a7248" },
    "yellow": { "name": "Cinnamon", "color": "#ffaa00" },
    "blue": { "name": "Nighttime", "color": "#2c6c9e" },
    "orange": { "name": "Clementine", "color": "#df5b06" },
  },
  "spring": {
    "red": { "name": "Veronika 👝", "color": "#feaeac" },
    "green": { "name": "Pear 🍐", "color": "#7dd17d" },
    "yellow": { "name": "Citrus 🍋", "color": "#efef7b" },
    "blue": { "name": "Soda 💧", "color": "#a9ccff" },
    "orange": { "name": "Shrimp 🦐", "color": "#ffb066" },
  },
}
```

If you have any questions, let me know!

christolis:
Awesome! A few questions though:
- I noticed that there are emojis only in the roles that belong in the `spring` category. Is this intentional?
- Should the announcements be different for when the roles have changed due to the new season and for when a new month is around, or should it be the same regardless of the scenario?
- I was thinking of creating a `nitro-roles.ts` file in `collections/` and storing all the different Nitro roles from the JSON above in there. Should this be the approach to take or is there a better one?

geisterfurz007:
1. Yes, at least that's how I found them in the list that Josh keeps.

2. Good point! Yeah, if you want, you can vary them (maybe also one like "Last chance for these colours this year" for the last month of a season?)!

3. I think such a file might make sense but I'd rather keep it together with the rest of the nitro stuff in the folder under programs; the corrections directory is an old relic that I would love to get rid of in the long run. Keeping features together in one folder makes the editing experience nicer, imo, because you don't need that many folders open at once.

src/programs/nitro-colors/roles-reset.ts
```diff
@@ -0,0 +1,94 @@
...
+            });
+          }
+        }
+
+        // Let Nitro boosters know about the new month's change!
+        await channel.send({
+          content: `${nitroBoosterRole} ${buildAnnouncementsMessage()} ${
+            season?.emoji
+          }`,
+        });
```
geisterfurz007, line 86: This is only sent for a new season; it should be sent for each new month.

src/programs/nitro-colors/roles-reset.ts
```diff
@@ -0,0 +1,94 @@
...
+  isNewSeason,
+} from "./common";
+
+export class RoleResetCron {
+  static init() {
+    const CRON =
+      process.env.NODE_ENV === "development" ? "* * * * *" : "0 0 1 * *";
+
+    // Schedule a cron task every month
+    cron.schedule(CRON, async () => {
```
geisterfurz007, line 19: This anonymous function is very large possibly making it hard to follow each step; instead of commenting each section of code describing what it does, you can extract them into smaller functions each with their respective name describing their function. This has several benefits including reducing indentation and making it easier to use editor features like jumping to each bit of code and collapsing unneeded sections.

geisterfurz007 (changes requested):

src/programs/nitro-colors/common.ts
```diff
@@ -0,0 +1,140 @@
...
+  let randomStr = (arr: string[]): string => arr[randomInt(arr.length)];
+
+  if (isNewSeason(month)) announcement.push(randomStr(announcements.newSeason));
+  else announcement.push(randomStr(announcements.normal));
+
+  if (isSeasonEnding(month))
+    announcement.push(" " + randomStr(announcements.endingSeason));
+
+  return announcement.join("");
+}
```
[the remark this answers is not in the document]
christolis: Good catch! I made them `const`. When writing the code, I thought it wouldn't matter initially, but on second thought, these need to be declared as constant.

src/programs/nitro-colors/roles-reset.ts
christolis: Fixed! 🚀

src/programs/nitro-colors/roles-reset.ts
christolis: Done! ✅

geisterfurz007 (approved):
LGTM!

This is great and will save us a lot of clicking and stuff! Thanks so much!

geisterfurz007:
Also sorry for all the delays in the reviews and thanks for your patience. I really appreciate it!

christolis:
> Also sorry for all the delays in the reviews and thanks for your patience. I really appreciate it!

No worries! I’m just really glad to be helping with getting these features made. I’d love to work on some more, if there are any! 🙌

PR #987, opened 2023-06-11
Sentry

### LINKED ISSUE #970: Set up Sentry
Occassionally the bot just crashes, however the bot restarts cause the errors to be lost.

Setting up sentry as a separate deployment (also for ytf) is probably a good idea at this scale to be able to pinpoint what is going on and to prevent further crashes in the future.

geisterfurz007:
Adds very rough and general tracking of errors to Sentry. Closes #970

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
   "dependencies": {
     "@discordjs/rest": "1.7.1",
     "@prisma/client": "4.15.0",
+    "@sentry/node": "^7.54.0",
     "axios": "1.4.0",
     "countries-and-timezones": "3.4.1",
     "country-code-emoji": "2.3.0",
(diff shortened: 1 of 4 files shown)


sklirg (approved):
LGTM. I would consider making it so _all_ events are sent to Sentry, instead of having to "hard-code" captureException into all possible failure cases, or make a convenience function to reduce boilerplate (thus making it simpler/less hassle, and increasing motivation/chances to use it in the future :))

geisterfurz007:
> LGTM. I would consider making it so _all_ events are sent to Sentry, instead of having to "hard-code" captureException into all possible failure cases, or make a convenience function to reduce boilerplate (thus making it simpler/less hassle, and increasing motivation/chances to use it in the future :))

The vast majority of errors will be caught in event-distribution/event-distribution.ts; the duplications in index.ts are only for remnants of old code that will have to be cleaned up eventually. Goal in the long run will definitely be to have one in the event-distribution in the core and one unhandledException one for everything that isn't caught properly (and would otherwise lead to a crash).

PR #992, opened 2023-06-17
fix: `UNKNOWN_ERROR` when a member is not in any group

sm-sami:
- fix: `UNKNOWN_ERROR`  which prevented the user from joining any group if they were not enrolled in at least one group.


geisterfurz007:
LGTM thanks a lot!

Issue #1031, opened 2023-07-25
Feature: Tag people when they're let in a voice room


joshtrivedi:
Right now the person who knocks a voice room is mentioned in Room chat with message "Hey X, Y wants to join your room", and after Y is let in, It only says "Alright, I made room for them!" but Y has no idea if they've been let in (the previous version used to tell you X let you in, or X didn't respond.

geisterfurz007:
Yeah, this should be possible! Currently we only defer the message of the result; instead we could immediately reply and use `followUp` to send another message that contains the ping once the result is available.

joshtrivedi:
wow this is old

PR #1032, opened 2023-07-26
Feature: Ping knocker once they are let in

christolis:
This PR attempts to implement issue #1031.


geisterfurz007:
@joshtrivedi can you try to get YesBot to run and see if this is what you had in mind?

geisterfurz007 (approved):
LGTM!

Issue #1379, opened 2024-08-02
Prevent bot from reacting to Discord native polls


julisod:
In the polls channel, the bot is designed to react to messages to create polls. With Discord's new native polls, that is not necessary.

The task would be to figure out if there's a way for the bot to recognise when the message is a native poll and to avoid reacting.

Below is a picture of what happens currently.

![image](https://github.com/user-attachments/assets/d6277cf1-dd7f-4025-84ee-bb9d6d1c9cdc)

julisod:
I'm planning on taking a look at this soon.

geisterfurz007:
Sounds good, thanks! I hope the contributing guide is still up-to-date. If you have any issues setting up YesBot, let me know!

If you are taking a look at this, could you also check if you can have YesBot avoid reactions on the system-messages that announce a finished poll?

PR #1388, opened 2024-08-05
fix: ignore Discord's polls

julisod:
- don't react to Discord's native polls
- don't react to poll results (or any other system messages) in polls
- fix the bug where you could not use custom emojis (that are on the server) when creating polls

#1379


geisterfurz007 (approved):
LGTM! Thanks a bunch!

